PR #15, opened 2015-08-27
Check version before each build

### LINKED ISSUE #10: Start using gitflow and semver release tagging
Sorry to feel so corporate, but as more folks start to use the app, some structure will help master be unbroken :innocent:

loudnate:
@ps2 This will simply halt the build with the error printed by the script. 
Because this is running as a pre-build phase, it's legal to modify source files here if we wanted to change the script to write the version numbers out to each location, choosing one as a source-of-truth.

Fixes #10


ps2:
Updated the script to actually modify CFBundleVersion and podspec version.

Issue #20, opened 2015-08-31
Set custom name for individual RileyLink


ps2:
Need to be able to assign a custom name, to distinguish between multiple RileyLinks.  This is already supported at the bluetooth level.

PR #21, opened 2015-09-03
Rename Rileylinks

ps2:
Need to figure out why peripheralDidUpdateName is not being called.  Likely an issue in the bgscript


loudnate:
Awesome!

loudnate:
Since you're renaming some enum values anyway, worth changing them to follow the cocoa naming and case conventions? And use the NS_ENUM macro? The Obj-C modernization tool can do the bulk of this for you.

ps2:
This will need the latest updates to the ble113 firmware.

RileyLink/RileyLinkBLEDevice.h
```diff
@@ -11,9 +11,9 @@
 
 
 typedef NS_ENUM(NSUInteger, RileyLinkState) {
-  RILEY_LINK_STATE_CONNECTING,
-  RILEY_LINK_STATE_CONNECTED,
-  RILEY_LINK_STATE_DISCONNECTED
+  RLStateConnecting,
+  RLStateConnected,
+  RLStateDisconnected
 };
```
loudnate, line 17: So the naming conventions state that each type should begin with the enum type name, e.g. `RileyLinkStateConnecting`. This automatically bridges to swift enums as `case Connecting`

loudnate:
LGTM.

RileyLink/RileyLinkBLEDevice.h
ps2, line 17: Ah, thanks.

ps2:
Fixes #20

PR #30, opened 2015-09-25
Simulate sensor

ps2:
Add interface to simulate enlite sensor data.  Should be able to specify sensor id, glucose trend, and have it run in a scheduled manner.


ps2:
Created against wrong destination branch.

PR #32, opened 2015-09-25
adding sensor simulator; work in progress

ps2:
Add interface to simulate enlite sensor data. Should be able to specify sensor id, glucose trend, and have it run in a scheduled manner.


ps2:
Closing this, as too much time/code has changed for it to be really of much use now; we'd do it differently nowadays.

PR #39, opened 2015-11-21
Pump history

### LINKED ISSUE #37: Poll Pump History
Poll pump history, and extract treatments, meter bg, etc.

ps2:
Currently polling page 0, and decoding into types; specific fields are not being decoded yet.

Fixes #37


loudnate:
:thumbsup:

Can you give more detail on why you needed to add the extra check to the NSNotification response? My assumption is that the serial queue would prevent any mismatch.

ps2:
So the sequence of events I saw was as follows:

Lets say I have two operations in the queue for the same pump id and packet type (0x80); they're both subscribed to RILEYLINK_EVENT_PACKET_RECEIVED. A response packet comes in and is scheduled to be sent to both of them.  The deviceDidReceivePacket callback for the first is called, finishAfterWaiting is called, the state is transitioned to MessageSendStateFinished, and isFinished is set to true, which lets the next operation start; it sends (queues) its packet and transitions to the MessageSendStateWaitingForReponse state.

All of that happens before the deviceDidReceivePacket callback for the second operation is called. When it is, it checks the pumpid and message type, and they match up, so it thinks it has a response for its outgoing packet.

If you remove the condition I added in deviceDidReceivePacket, and put a check to see if sentAt is after rxPacket.capturedAt, you'll be able to catch it happening by doing a history dump.

loudnate:
it sounds like we should move the NSNotificationCenter subscription to the
point where we're waiting for response instead of in `init`, then.

ps2:
Are we assured it's not possible to receive notifications that were sent before we subscribed? I could imagine an implementation of the distribution process that does not guarantee this.  If can make that assumption, I think I'd agree and go with the delayed subscription tack.

loudnate:
No, I'm not assured, but if you don't specify a queue during subscription, the notifications are received on the same queue as they were posted, so I'd bet deferring the subscription would work. 

The extra check isn't hurting anything, though, so I'm not seeing any reason to remove it even if we moved the subscription point.

Issue #42, opened 2015-12-28
MessageSendOperation main vs start


ps2:
Review moving the code in MessageSendOperation -main back to -start. I had moved it when I encountered issues cancelling operations before they started, but that might not be the right fix.

ps2:
Maybe related:

```
2015-12-28 21:22:01.612 RileyLink[1180:481314] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSOperationInternal _start:]: something is trying to start the receiver simultaneously from more than one thread'
*** First throw call stack:
(0x1840b9900 0x183727f80 0x1840b9848 0x1849a320c 0x184a62728 0x1001edbb0 0x1001f3658 0x184070bb0 0x18406ea18 0x183f9d680 0x1854ac088 0x188e14d90 0x1001033b4 0x183b3e8b8)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

This was in a version where the code is in -main

loudnate:
Won't Fix!

PR #44, opened 2016-01-06
Rfspy

ps2:
A lot of changes here.  Too many.   Probably should have broken this up more, but here it is.  This change will make the app depend on having the subg_rfspy firmware on the cc1110, and the ble113_rfspy firmware on the ble113.


ps2:
Fixes #42

ps2:
closes #37

PR #49, opened 2016-01-25
Wip/synchronous commands

ps2:
Big rewrite of command handling.  Now we have "sessions".  Each session is serialized so only one has access to the RL device at any time.  Timeouts and idle are working also.


ps2:
Ugh, this is against master.

PR #57, opened 2016-02-28
Persist invoking idle through disconnects

loudnate:
Let me know if I'm misunderstanding the concept of the idle trigger here.

### DIFF
--- a/RileyLink/RileyLinkBLEDevice.m
+++ b/RileyLink/RileyLinkBLEDevice.m
@@ -177,7 +177,22 @@
   return rval;
 }
 
-- (void) didDisconnect:(NSError*)error {
+- (void)connectionStateDidChange:(NSError *)error
+{
+  switch (self.peripheral.state) {
+    case CBPeripheralStateConnected:
+      if (idleListeningEnabled) {
+        [self onIdle];
+      }
+      break;
+    case CBPeripheralStateDisconnected:
+      runningIdle = NO;
+      runningSession = NO;
+      break;
+    case CBPeripheralStateConnecting:
+    case CBPeripheralStateDisconnecting:
+      break;
+  }
 }
 
 - (void)setCharacteristicsFromService:(CBService *)service {
@@ -427,6 +442,7 @@
 
 - (void) disableIdleListening {
   idleListeningEnabled = NO;
+  runningIdle = NO;
 }
 
 - (void) handleIdleListenerResponse:(NSData *)response {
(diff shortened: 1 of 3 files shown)


RileyLink/RileyLinkBLEDevice.m
```diff
@@ -177,7 +177,22 @@ - (RileyLinkState) state {
   return rval;
 }
 
-- (void) didDisconnect:(NSError*)error {
+- (void)connectionStateDidChange:(NSError *)error
+{
+  switch (self.peripheral.state) {
```
loudnate, line 182: Yep, I changed my global settings to 2 spaces just for RLIOS dev.

ps2, line 182: :)

ps2:
I _think_ this is ok. Yes we could lose an idle response while disconnected.  Riley goes in and out of range all the time, and I haven't noticed I'm missing data because of it, but we're on dex right now, so it wouldn't be as visible.  The only danger to this is getting back an extra response from an interrupted idle when you reconnect and issue a rx on top of the rx that was already running.

loudnate:
That's a good point. Should we guard `onIdle` with `&& runningIdle == NO` then?

``` objc
- (void) onIdle {
   if (idleListeningEnabled) {
     runningIdle = YES;
     NSLog(@"Starting idle RX");
     GetPacketCmd *cmd = [[GetPacketCmd alloc] init];
     cmd.listenChannel = idleListenChannel;
     cmd.timeoutMS = 60 * 1000;
     [self issueCommand:cmd];
   }
 }
```

loudnate:
We might also want to dispatch to the main queue here before accessing `idleListeningEnabled`, since that's where all the BLE delegate calls are running.

``` objc
- (void) enableIdleListeningOnChannel:(uint8_t)channel
```

ps2:
No, that's pretty much already guarded for elsewhere (we could guard inside there as well or alternately).  
You mark runningIdle = NO when we disconnect, but it's possible that the rx command is still running on the RL when we reconnect, and issue another RX.

loudnate:
Will a second `GetPacketCmd` interrupt the first?

ps2:
Yes, and will generate a SubgRfspyErrorCmdInterrupted response.

loudnate:
Ah, so that cycle of interruption would continue then. What about sending a reset command on reconnect, if `runningIdle == YES`?

ps2:
Then we lose tuning info.  If we do observe it getting into a loop (I'm not sure it will), then we can look at catching the CmdInterrupted earlier, and not leaving idle state.

loudnate:
I can try and force the contrived case and see how it recovers.

PR #59, opened 2016-03-19
Increase timeout for listen commands

loudnate:


loudnate:
A version of this change is already merged into `dev`.

PR #63, opened 2016-04-09
Reconcile with naterade-ios/MinimedKit

loudnate:


ps2:
Looks great.  Thanks!

PR #64, opened 2016-04-09
Removing podspec, bumping version

loudnate:


loudnate:
@ps2 you'll need to turn on travis support in your account.

loudnate:
Also, any time you'd like to drop support for earlier iOS versions, I'd be down.

PR #69, opened 2016-04-10
Pump ops cleanup

loudnate:
my end goal is to move the operations in Naterade (bolus, change time, change temp basal, read temp basal) to `RileyLinkKit`.

### DIFF
--- a/RileyLinkKit/PumpOpsSynchronous.swift
+++ b/RileyLinkKit/PumpOpsSynchronous.swift
@@ -32,7 +32,7 @@ public class PumpOpsSynchronous: NSObject {
   }
   
   private func makePumpMessage(messageType: MessageType, body: MessageBody) -> PumpMessage {
-    return PumpMessage.init(packetType: .Carelink, address: pump.pumpID, messageType: messageType, messageBody: body)
+    return PumpMessage(packetType: .Carelink, address: pump.pumpID, messageType: messageType, messageBody: body)
   }
   
   public func sendAndListen(msg: PumpMessage, timeoutMS: UInt16 = standardPumpResponseWindow, repeatCount: UInt8 = 0, msBetweenPackets: UInt8 = 0, retryCount: UInt8 = 3) -> PumpMessage? {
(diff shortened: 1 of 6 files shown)


RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -15,6 +15,7 @@ public enum PumpCommsError: ErrorType {
   case RFCommsFailure(String)
   case UnknownPumpModel
   case RileyLinkTimeout
+  case UnknownResponse
```
loudnate, line 18: This file is best viewed side-by-side (Diff Options)

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -157,8 +149,12 @@ public class PumpOpsSynchronous: NSObject {
     
     let frequencies = [916.55, 916.60, 916.65, 916.70, 916.75, 916.80]
     var results = FrequencyScanResults()
-    
-    defaultWake()
+
+    do {
+      try wakeup()
+    } catch {
+      // Continue anyway
```
loudnate, line 156: I believe this was your intention in the original implementation?

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -199,12 +195,14 @@ public class PumpOpsSynchronous: NSObject {
...
+
+    do {
+      try wakeup()
+    } catch _ as PumpCommsError {
       try scanForPump()
     }
-    
-    guard let pumpModelStr = getPumpModel() else {
+
+    guard let pumpModelStr = try? getPumpModel() else {
```
loudnate, line 205: `try?` isn't a great pattern in all cases because it ignores what the underlying exception was. Rather than re-think any logic, my goal was to keep the logic exactly the same but adjust the implementation.

They say that every time you write a `!`, Apple deletes a Radar.

RileyLinkKit/PumpOpsSynchronous.swift
ps2, line 156: Yes. If the frequency is too far off, we won't hear the pump's response, but it probably did wake. So we want to continue with scanning.

ps2:
These are changes I'd like to test, but I think it's ok to pull into dev; at this point dev needs a good regression test anyways.

PR #71, opened 2016-04-14
RileyLinkKit: High-level device interface

loudnate:
This is an nearly-exact port of the interfaces I've been using in Naterade since August of last year. It provides a high-level interface for working with RileyLink devices and running pump commands.

Here's an example of where my `LoopDataManager` [enacts its recommended temp basal](https://github.com/loudnate/naterade-ios/blob/rileylinkkit/Naterade/Managers/LoopDataManager.swift#L432-L442)


ps2:
Is this something that should eventually just be part of the RLBLE classes, or do you see it staying as a separate layer?

ps2:
Ah, guessing keep it separate as it depends on MinimedKit.   I can see the utility it provides.  LGTM

loudnate:
Yeah. I think we instruct folks to use the `RileyLinkDeviceManager` class if they're working with Minimed pumps, and update the RileyLink example app to follow. 

I also plan to normalize both of our device utility interfaces into a single root view controller, and include them in `RileyLinkKit`.

PR #72, opened 2016-04-19
Updating CFBundleVersion (2) and CFBundleShortVersionString (0.2.0)

loudnate:
Use `agvtool` to easily update these across all projects in the future:
https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/agvtool.1.html


ps2:
Nice; didn't know about that tool.  LGTM

loudnate:
Yeah I only recently found out about it, but it's been around forever. Easier than plistbuddy.

PR #78, opened 2016-04-26
Surface the firmware version; remove the UIAlertView

loudnate:


ps2:
At some point I'll want to add a notice back for users, but considering the issues you're having, I think your changes are better for now.

loudnate:
I surfaced the status and version string to enable that from a view controller. 
`UIAlertView` is deprecated, and it only made this code possible because it used a separate UIWindow to display the dialog. Its replacement, `UIAlertController`, needs to be presented on the app's view controller hierarchy and therefore won't ever happen from `RileyLinkBLEDevice` anyway.

loudnate:
I'm going to merge this down. I still plan to combine the feature set of both the [Naterade](https://github.com/loudnate/naterade-ios/blob/master/Naterade/View%20Controllers/RileyLinkDeviceTableViewController.swift) and [RileyLink](https://github.com/ps2/rileylink_ios/blob/dev/RileyLink/RileyLinkDeviceViewController.h) versions of the `DeviceViewController` into one, and include that in `RileyLinkKit`. That view controller would be responsible for warning the user about a bad firmware version state.

PR #79, opened 2016-04-26
Treat a scan with 0 successes as a failure

loudnate:


ps2:
Hmm, wonder why the test failed.

loudnate:
Yeah I just now got a chance to look into this. I was waiting for [xctool](https://github.com/facebook/xctool/releases) to fix their Xcode 7.3 support, and it looks like that's released now... the unsanitized log output is awful.

loudnate:
Looks like the failure was transient; it's passed now.

PR #83, opened 2016-04-29
RILEYLINK_EVENT_DEVICE_CONNECTED isn't always posted when restoring

loudnate:


loudnate:
It sounds like this also fixes a crash in the RL app for the start-with-timeout condition.

PR #91, opened 2016-05-18
Device views

ps2:


RileyLinkKit/Extensions/NSDate.swift
```diff
@@ -0,0 +1,20 @@
...
+//  RileyLink
+//
+//  Created by Pete Schwamb on 5/18/16.
+//  Copyright © 2016 Pete Schwamb. All rights reserved.
+//
+
+import Foundation
+
+
+public func ==(lhs: NSDate, rhs: NSDate) -> Bool {
```
loudnate, line 11: Can we keep these `internal` to the framework?

RileyLinkKit/ViewControllers/MySentryPairViewController.swift
```diff
@@ -179,7 +177,7 @@ class MySentryPairViewController: UIViewController, UITextFieldDelegate {
         
         if let data = packet.data, msg = PumpMessage(rxData: data) {
             if msg.packetType == PacketType.MySentry &&
-                msg.address.hexadecimalString == Config.sharedInstance().pumpID {
+                msg.address.hexadecimalString == device.pumpState!.pumpID {
```
loudnate: Is `pumpState` guaranteed to be non-nil to run this view controller? I'm not sure if that's the case today with Naterade.

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
```diff
@@ -0,0 +1,353 @@
...
+        super.viewDidLoad()
+
+        title = device.name
+    }
+
+    public override func viewWillAppear(animated: Bool) {
+        super.viewWillAppear(animated)
+
+        if appeared {
+            tableView.reloadData()
```
loudnate, line 28: I don't love this. But we can clean it up later.

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
```diff
@@ -0,0 +1,353 @@
...
+                let localTimeZone = NSTimeZone.defaultTimeZone()
+                let localTimeZoneName = localTimeZone.abbreviation ?? localTimeZone.name
+
+                if let pumpTimeZone = device.pumpState?.timeZone {
+                    let timeZoneDiff = NSTimeInterval(pumpTimeZone.secondsFromGMT - localTimeZone.secondsFromGMT)
+                    let formatter = NSDateComponentsFormatter()
+                    formatter.allowedUnits = [.Hour, .Minute]
+                    let diffString = timeZoneDiff != 0 ? formatter.stringFromTimeInterval(abs(timeZoneDiff)) ?? String(abs(timeZoneDiff)) : ""
+
+                    cell.detailTextLabel?.text = String(format: NSLocalizedString("%1$@%2$@%3$@", comment: "The format string for displaying an offset from a time zone: (1: GMT)(2: -)(3: 4:00)"), localTimeZoneName, timeZoneDiff != 0 ? (timeZoneDiff < 0 ? "-" : "+") : "", diffString)
```
loudnate, line 184: Unfortunately `NSDateComponentsFormatter` doesn't have the ability to explicitly set `positiveFormat` and `negativeformat` like `NSNumberFormatter` does, which could clean up the "+"/"-" bit.

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
```diff
@@ -0,0 +1,353 @@
...
+                } else {
+                    cell.detailTextLabel?.text = localTimeZoneName
+                }
+            case .MySentryPair:
+                cell.textLabel?.text = "MySentry Pair"
```
loudnate: Should we use `NSLocalizedString` to be consistent? 

Gotta get those RileyLinks all over the globe!

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
```diff
@@ -0,0 +1,353 @@
...
+                    }
+
+                    return "Changing time..."
+                }
+
+                vc.title = "Change Time"
+
+                self.showViewController(vc, sender: indexPath)
+            case .MySentryPair:
+                let vc = self.storyboard!.instantiateViewControllerWithIdentifier("mySentryPair") as! MySentryPairViewController
```
loudnate: Set MySentryPairViewController to extend from `IdentifiableClass` and you can use `MySentryPairViewController.className` instead of the arbitrary string for the identifier. (Make sure to change it in the storyboard too)

loudnate: Also I think we might be able to get rid of the custom MySentryPairViewController altogether since it turns out the MySentry ID didn't end up being useful at all. `CommandResponseViewController` could start with the pairing instructions as it starts pinging.

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
```diff
@@ -0,0 +1,353 @@
...
+                
+                vc.device = device
+                
+                vc.title = "MySentry Pair"
+                
+                self.showViewController(vc, sender: indexPath)
+            case .DumpHistory:
+                let vc = CommandResponseViewController { [unowned self] (completionHandler) -> String in
+                    let calendar = NSCalendar.currentCalendar()
```
loudnate: This should be `NSCalendar(identifier: NSCalendarIdentifierGregorian)`, since users could be using any calendar on their phone.

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
```diff
@@ -0,0 +1,353 @@
...
+                    let oneDayAgo = calendar.dateByAddingUnit(.Day, value: -1, toDate: NSDate(), options: [])
+                    self.device.ops?.getHistoryEventsSinceDate(oneDayAgo!) { (response) -> Void in
```
loudnate, line 300: similar question as above, can ops be nil here? We should either force-unwrap or add a `guard` statement to line 296.5 so we don't show the view controller.

RileyLinkKit/Extensions/NSDate.swift
ps2, line 11: I brought this in to handle some date comparisons that you had written, but I couldn't find out where you had implemented this.  

When I try to make the funcs or the extension itself private, the compiler doesn't accept it.

RileyLinkKit/ViewControllers/MySentryPairViewController.swift
ps2: good catch

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
ps2, line 28: The 'appeared' var?

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
ps2: Yes, good idea

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
loudnate, line 28: No, the reloadData on every appearance.

RileyLinkKit/Extensions/NSDate.swift
```diff
@@ -0,0 +1,19 @@
...
+//  NSDate.swift
+//  RileyLink
+//
+//  Created by Pete Schwamb on 5/18/16.
+//  Copyright © 2016 Pete Schwamb. All rights reserved.
+//
+
+import Foundation
+
+public func ==(lhs: NSDate, rhs: NSDate) -> Bool {
```
loudnate, line 11: There's a third scope (default) which is `internal`. It's availabile only to the framework.

ps2, line 11: That works for the ==, but not the <

loudnate, line 11: ah, right, because the protocol is public. I'm concerned that two frameworks with the same public extension would conflict...

PR #93, opened 2016-05-21
Simple nav

ps2:
get rid of nightscout webview, as it mostly got in the way, bring in a simpler settings setup, ala naterade.


RileyLink/TextFieldTableViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+
+import UIKit
+
+protocol TextFieldTableViewControllerDelegate: class {
+    func textFieldTableViewControllerDidEndEditing(controller: TextFieldTableViewController)
+}
+
+
+class TextFieldTableViewController: UITableViewController, IdentifiableClass, UITextFieldDelegate {
```
loudnate, line 16: I had been meaning to move this into LoopKitUI, since it's useful boilerplate :)

ps2, line 16: Yes, that would be nice!

PR #96, opened 2016-05-22
Organization and conflict resolution

loudnate:


RileyLink/RileyLinkDeviceTableViewCell.swift
```diff
@@ -1,51 +1,13 @@
 //
 //  RileyLinkDeviceTableViewCell.swift
-//  Naterade
+//  RileyLink
```
loudnate, line 3: Sorry, this diff got weird because I moved the file out of `RileyLink/` into the framework's directory, then re-added a file in its place which declared `IdentifiableClass` conformance (since that's really just a storyboard convenience, it should stay in the target with the storyboard)

RileyLinkKit/Extensions/NSDate.swift
```diff
@@ -8,12 +8,12 @@
...
 }
 
-public func <(lhs: NSDate, rhs: NSDate) -> Bool {
+// MARK: - Comparable
+
+func <(lhs: NSDate, rhs: NSDate) -> Bool {
     return lhs.compare(rhs) == .OrderedAscending
 }
-
-extension NSDate: Comparable { }
```
loudnate, line 19: Since the only code that leverages this function was comparing `NSDate` explicitly, it didn't need to be declared `Comparable`. I actually ended up removing the only use of this function below, anyway.

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
```diff
@@ -145,7 +145,7 @@ public class RileyLinkDeviceTableViewController: UITableViewController {
                 }
             case .Awake:
                 switch device.pumpState?.awakeUntil {
-                case let until? where until < NSDate():
+                case let until? where until.timeIntervalSinceNow < 0:
```
loudnate, line 148: This was just a silly original implementation by me. No sense instantiating a new `NSDate` object for this.

RileyLinkKit/ViewControllers/RileyLinkDeviceTableViewController.swift
```diff
@@ -286,13 +286,13 @@ public class RileyLinkDeviceTableViewController: UITableViewController {
...
                 self.showViewController(vc, sender: indexPath)
             case .MySentryPair:
-                let vc = self.storyboard!.instantiateViewControllerWithIdentifier(MySentryPairViewController.className) as! MySentryPairViewController
-                
-                vc.device = device
-                
-                vc.title = NSLocalizedString("MySentry Pair", comment: "Title of screen for pairing with MySentry.")
-                
-                self.showViewController(vc, sender: indexPath)
+                if let vc = self.storyboard?.instantiateViewControllerWithIdentifier(MySentryPairViewController.className) as? MySentryPairViewController {
```
loudnate, line 289: I'm going to remove the storyboard dependency from `RileyLinkKit` entirely, but for now I'd rather this fail silently for users without identical storyboard setups than crash.
If we felt like we really needed a storyboard/nib, then we should create a separate one and package it with the framework.

RileyLinkKit/Views/RileyLinkDeviceTableViewCell.swift
```diff
@@ -0,0 +1,51 @@
+//
+//  RileyLinkDeviceTableViewCell.swift
```
loudnate, line 2: This was just a `git mv` until I re-added a file in its place.

PR #97, opened 2016-05-22
Allow scan frequencies to be configured

ps2:


loudnate:
shall we take the last tune date out of RileyLinkDevice then? I'm not sure "last tune" is a Pump property vs a RileyLink property. It's possible to use the same pump state but change rileylinks, no?

ps2:
You're right; it belongs on RLDevice

RileyLinkKit/PumpState.swift
```diff
@@ -23,8 +23,10 @@ public class PumpState {
     /// If the value type does not conform to AnyObject, a raw representation will be provided instead.
     public static let ValueChangeOldKey = "com.rileylink.RileyLinkKit.PumpState.ValueChangeOldKey"
     
-    public let pumpID: String
+    public var scanFrequencies = [916.45, 916.50, 916.55, 916.60, 916.65, 916.70, 916.75, 916.80]
```
loudnate, line 26: I worry about any public mutable array anywhere :) But let's see where this goes.

loudnate:
LGTM

PR #113, opened 2016-06-09
Retry wakeup — more consistent communications for x22 pumps

kamens:
(Apologies if I should base this PR off of a branch other than dev. Lemme know the process you'd prefer.)

As discussed in gitter/slack, this PR adds retry logic to the initial short wakeup attempt, which has been verified by me and aemazaheri to increase likelihood of RL<=>pump communication success.

You and @loudnate know far more than I do about _why_ that is the case ;). Happy to add more findings/context from my experience here if helpful.

### DIFF
--- a/RileyLinkKit/PumpOpsSynchronous.swift
+++ b/RileyLinkKit/PumpOpsSynchronous.swift
@@ -86,7 +86,6 @@ class PumpOpsSynchronous {
                     lastError = PumpCommsError.UnknownResponse("Wakeup shortResponse: \(shortResponse.txData)")
                 }
             } catch let error {
-                print("Wakeup failure on attempt #\(attempt): \(error)")
                 lastError = error
             }
         }
@@ -102,7 +101,7 @@ class PumpOpsSynchronous {
             return
         }
         
-        try attemptShortWakeUp();
+        try attemptShortWakeUp()
         
         let longPowerMessage = makePumpMessage(.PowerOn, body: PowerOnCarelinkMessageBody(duration: duration))
         let longResponse = try sendAndListen(longPowerMessage)
--- a/RileyLinkKit/PumpOpsSynchronous.swift
+++ b/RileyLinkKit/PumpOpsSynchronous.swift
@@ -62,18 +62,47 @@ class PumpOpsSynchronous {
         return message
     }
     
+    /**
+     Attempts to send initial short wakeup message that kicks off the wakeup process.
+
+     Makes multiple attempts to send this message in order to work around issue w/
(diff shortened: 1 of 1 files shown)


RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -62,18 +62,47 @@ class PumpOpsSynchronous {
...
+                let shortResponse = try sendAndListen(shortPowerMessage, timeoutMS: 15000, repeatCount: 255, msBetweenPackets: 0, retryCount: 0)
+
+                if shortResponse.messageType == .PumpAck {
+                    // Pump successfully received and responded to short wakeup message!
+                    return
+                } else {
+                    lastError = PumpCommsError.UnknownResponse("Wakeup shortResponse: \(shortResponse.txData)")
+                }
+            } catch let error {
+                print("Wakeup failure on attempt #\(attempt): \(error)")
```
loudnate: This is probably unnecessary logging, since it's not something we previously logged.

Eventual stretch goal would be to remove all direct NSLog calls and let the framework's client choose how to handle logging, but that's a long way off.

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -62,18 +62,47 @@ class PumpOpsSynchronous {
...
         }
         
-        let shortPowerMessage = makePumpMessage(.PowerOn)
-        let shortResponse = try sendAndListen(shortPowerMessage, timeoutMS: 15000, repeatCount: 200, msBetweenPackets: 0, retryCount: 0)
-        
-        guard shortResponse.messageType == .PumpAck else {
-            throw PumpCommsError.UnknownResponse("Wakeup shortResponse: \(shortResponse.txData)")
-        }
-        NSLog("Pump acknowledged wakeup!")
+        try attemptShortWakeUp();
```
loudnate: unnecessary semicolon 😉

RileyLinkKit/PumpOpsSynchronous.swift
kamens: good call, fixed

RileyLinkKit/PumpOpsSynchronous.swift
kamens: thx, fixed :).

loudnate:
Changes look good to me. I want to wait for @ps2's blessing to merge.

ps2:
Looks good to me.  I wanted to try it out, but haven't found time.  I think we should go ahead and merge, since it's helping others.

PR #115, opened 2016-06-12
Cleanup toward history timezone reconciliation.

loudnate:
I'll probably submit a few of these as I get to them, which should keep reviewing manageable. 

The end goal is to use `PumpState.timeZone` + `ChangeTimePumpEvent` to assign a correct UTC date to each `PumpEvent`, which will maintain correctness during time zone changes by the phone, e.g. DST.

### DIFF
--- a/MinimedKit/PumpEvents/ChangeTimePumpEvent.swift
+++ b/MinimedKit/PumpEvents/ChangeTimePumpEvent.swift
@@ -8,9 +8,9 @@
 
 import Foundation
 
-public class ChangeTimePumpEvent: PumpEvent {
+public class ChangeTimePumpEvent: TimestampedPumpEvent {
     public let length: Int
-    let timestamp: NSDateComponents
+    public let timestamp: NSDateComponents
     
     public required init?(availableData: NSData, pumpModel: PumpModel) {
         length = 14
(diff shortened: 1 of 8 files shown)


MinimedKit/PumpEvents/ChangeTimePumpEvent.swift
```diff
@@ -8,9 +8,9 @@
 
 import Foundation
 
-public class ChangeTimePumpEvent: PumpEvent {
+public class ChangeTimePumpEvent: TimestampedPumpEvent {
```
loudnate, line 11: The majority of the `PumpEvent`-conforming classes also conform to `TimestampedPumpEvent`. Is there any reason they aren't declared as such?

ps2, line 11: You're right, they should be; I think it might only be UnabsorbedInsulinPumpEvent that doesn't have a timestamp?

ps2:
LGTM!

MinimedKit/PumpEvents/ChangeTimePumpEvent.swift
loudnate, line 11: I believe so, yeah, that and the midnight events (which only include YMD components)

PR #118, opened 2016-06-15
Swiftifying the NSDateComponents parsing

loudnate:
(stacked on #116)

PR #120, opened 2016-06-15
Using structs for PumpEvents

loudnate:
Stacked on #118


ps2:
LGTM.  Tested with NS Upload

PR #121, opened 2016-06-16
Reconciling PumpEvent timestamps to Dates using the pump's UTC offset

loudnate:
- Removes all uses of `localTimeZone` and `defaultTimeZone` from history parsing, instead using `pumpState.timeZone`

Note that this is stacked on #120


RileyLink.xcodeproj/project.pbxproj
```diff
@@ -1356,34 +1366,34 @@
 				TargetAttributes = {
 					430D64CA1CB855AB00FCA750 = {
 						CreatedOnToolsVersion = 7.3;
-						DevelopmentTeam = UY678SP37Q;
+						DevelopmentTeam = 57NRR26737;
```
loudnate: My bad, I'll revert these.

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -325,39 +325,35 @@ class PumpOpsSynchronous {
...
         
         let pumpModel = try getPumpModel()
-        
-        var pageNum = 0
-        var events = [PumpEvent]()
-        while pageNum < 16 {
+
+        var events = [TimestampedHistoryEvent]()
+
+        pages: for pageNum in 0..<16 {
```
loudnate, line 335: Here's some fun Swift: labeled control flows

RileyLink.xcodeproj/project.pbxproj
ps2: I have them reverted locally; if you want, I can commit.

RileyLinkKit/PumpOpsSynchronous.swift
ps2, line 335: now we're back to gotos. ;)

ps2:
NS uploading of temps and boluses appears to be working great.

RileyLinkKit/PumpOpsSynchronous.swift
loudnate, line 335: Hahahaha I suppose you're right! This was the only use case in 2 years I've found for them.

PR #122, opened 2016-06-17
Handling time change events during history UTC reconciliation

loudnate:
Before -> After

img_2520 img_2519


ps2:
I firsthand experienced the issue with using local phone time zone as the timezone for pump events when we few out here on friday, so looking forward to all this history goodness you've been cooking up!

This change looks good, also. 👍

loudnate:
Loop/MySentry Packers should have worked correctly with the time change, right?

ps2:
Yeah, that was working.  It was the timestamp on events that were uploaded to NS.  So maybe that’s still an issue; I’ll need to check.

loudnate:
I didn't change how MySentry packet timestamps were handled in the RL app. I haven't gone through and made sure that PumpState.timezone is being serialized to NSUserDefaults correctly there either, which would be a prerequisite.

PR #124, opened 2016-06-19
Handling the mutability of bolus history entries

loudnate:


ps2:
👍

loudnate:
Was I forgetting about any other potentially mutable PumpEvent entries?

PR #125, opened 2016-06-21
Version 0.4.0

loudnate:
Some x22 users have been complaining of timeouts with Loop, and I forgot we hadn't released @kamens's patch.

I'll update the PR with a version string in a minute; for now I'm opening this to audit for compatibility.


RileyLinkKit/PumpOps.swift
```diff
@@ -96,7 +96,7 @@ public class PumpOps {
         }
     }
     
-    public func getHistoryEventsSinceDate(startDate: NSDate, completion: (Either<(events: [PumpEvent], pumpModel: PumpModel), ErrorType>) -> Void) {
+    public func getHistoryEventsSinceDate(startDate: NSDate, completion: (Either<(events: [TimestampedHistoryEvent], pumpModel: PumpModel), ErrorType>) -> Void) {
```
loudnate, line 99: Here's our breaking API change. Going to call this `0.4.0`.

ps2:
Were those reports for "RileyLinkTimeout" errors? Those should not be caused by pump wakeup issues. RileyLinkTimeout errors indicate something wrong at the ble/spi comms layers

ps2:
LGTM

loudnate:
Re: `RileyLinkTimeout`, I'm not sure. @kamens @mddub can you elaborate?

mddub:
Without the `attemptShortWakeUp` fix I had long periods of being unable to manually run pump model or poll reservoir commands from Loop's "device settings" pane. I forget the exact error message, but the command would come back with something like `Sent <a70000000>, received <>` (where `000000` is the serial). Also don't quote me on the `a7`.

Here's what I would see in my Loop error log when the closed loop was failing:
https://gist.github.com/mddub/ad71d7fb77ec120587c0b7418008f89c

PR #126, opened 2016-06-21
Adding Questionable3b; handling pumps with empty history

loudnate:
Tested history fetching with a pump that consistently refused to vend page 1 due to it being empty.

This change carries a risk that a secondary page fails for a temporary reason but we return success with just the earlier pages, and never know to come back. Thoughts?


RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -335,7 +335,17 @@ class PumpOpsSynchronous {
 
         pages: for pageNum in 0..<16 {
             NSLog("Fetching page %d", pageNum)
-            let pageData = try getHistoryPage(pageNum)
+            let pageData: NSData
+
+            do {
+                pageData = try getHistoryPage(pageNum)
+            } catch let error as PumpCommsError {
+                if pageNum > 0 {
```
ps2: What error was actually thrown in the case of the pump that was refusing to vend page 1?

loudnate: `throw PumpCommsError.RFCommsFailure("Pump not responding to GetHistory command")`

loudnate: Once we tell it we want page 1, it either didn't respond or responded with a different opcode. (I didn't check, but I should do that tonight)

ps2: Yeah, that's what I was wondering; I left my empty history 522 at home, so I can't test.

loudnate: So I got back the following (tried it 3 times for good measure): `a7xxxxxx 15 0d 80`

At this point we don't have a `MessageType` entry for `0x15`. I don't know what the `0d` refers to, since this fails when trying to read page 1. 

I'll change this implementation to accept the `0x15` message type and halt further reading.

ps2: Cool, we found the "no more pages" message!

MinimedKit/MessageType.swift
```diff
@@ -15,6 +15,7 @@ public enum MessageType: UInt8 {
     case PumpBackfill           = 0x08
     case FindDevice             = 0x09
     case DeviceLink             = 0x0A
+    case EmptyHistoryPage       = 0x15
```
loudnate, line 18: Here's our new message,

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -15,7 +15,8 @@ public enum PumpCommsError: ErrorType {
     case RFCommsFailure(String)
     case UnknownPumpModel
     case RileyLinkTimeout
-    case UnknownResponse(String)
+    case UnknownResponse(rx: NSData?, during: String?)
+    case UnexpectedResponse(PumpMessage, from: PumpMessage)
```
loudnate, line 19: This clarifies a lot of `UnknownResponse` errors, and removes hand-crafted strings, which will make it easier to localize the user-facing portion if necessary.

loudnate:
This is ready for re-review.

RileyLinkKit/PumpOpsSynchronous.swift
ps2, line 19: 👍

ps2:
LGTM!

PR #127, opened 2016-06-22
Timezone handling for pump status uploads

ps2:
Use configured pump timezone to interpret dates from mysentry packets when uploading to Nightscout


ps2:
wrong target branch

PR #128, opened 2016-06-22
Timezone handling for pump status uploads

ps2:
Use configured pump timezone to interpret dates from mysentry packets when uploading to Nightscout

### DIFF
--- a/NightscoutUploadKit/NightscoutUploader.swift
+++ b/NightscoutUploadKit/NightscoutUploader.swift
@@ -142,12 +142,17 @@ public class NightscoutUploader: NSObject {
             nsStatus["uploader"] = ["battery":uploaderDevice.batteryLevel * 100]
         }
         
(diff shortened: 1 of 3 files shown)


NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -142,12 +142,17 @@ public class NightscoutUploader: NSObject {
             nsStatus["uploader"] = ["battery":uploaderDevice.batteryLevel * 100]
         }
         
-        let pumpDate = TimeFormat.timestampStr(status.pumpDateComponents)
+        guard let pumpDate = status.pumpDateComponents.date else {
+            NSLog("Pump date not set (or timezone missing) in pump status message!")
```
loudnate, line 146: Maybe we can circle back and `throw` instead of logging here?

loudnate:
LGTM!

PR #130, opened 2016-06-23
Add async error handling for nightscout upload errors

ps2:
Surface nightscout upload errors to devicedatamanager.

### DIFF
--- a/NightscoutUploadKit/NightscoutUploader.swift
+++ b/NightscoutUploadKit/NightscoutUploader.swift
@@ -10,6 +10,14 @@ import UIKit
 import MinimedKit
 import Crypto
 
+public enum UploadError: ErrorType {
+    case MissingAPISecret
+    case MissingNightscoutURL
+    case InvalidNightscoutURL(String)
+    case HTTPError(status: Int, body: String)
+    case MissingTimezone
+}
+
 public class NightscoutUploader: NSObject {
 
     enum DexcomSensorError: UInt8 {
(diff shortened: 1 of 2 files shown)


NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -311,7 +331,7 @@ public class NightscoutUploader: NSObject {
         treatmentsQueue =  [AnyObject]()
         uploadToNS(inFlight, endpoint: defaultNightscoutTreatmentPath) { (error) in
             if error != nil {
```
loudnate:
``` swift
if let error = error {
   self.errorHandler?(error: error, context: "...")
   // ...
}
```

(also above on line 322)

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -287,7 +307,7 @@ public class NightscoutUploader: NSObject {
         deviceStatuses =  [AnyObject]()
         uploadToNS(inFlight, endpoint: defaultNightscoutDeviceStatusPath) { (error) in
             if error != nil {
```
loudnate: Use optional binding here (see below)

RileyLink/DeviceDataManager.swift
```diff
@@ -243,6 +243,9 @@ class DeviceDataManager {
         nightscoutUploader = NightscoutUploader()
         nightscoutUploader.siteURL = nightscoutURL
         nightscoutUploader.APISecret = nightscoutAPISecret
+        nightscoutUploader.errorHandler = { (error: ErrorType, context: String) -> Void in
+            NSLog("Error %s, while %s", String(error), context)
```
loudnate: Does `%s` work here? In Obj-C you'd need to use `%@`.

loudnate: actually, better to just use `print("Error \(error), while \(context)")`

loudnate:
LGTM, thanks for those changes!

PR #131, opened 2016-06-23
Track timestamp of last successfully uploaded treatment from pump

ps2:
Used to specify how far back in time we need to look for events from the pump


NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -31,11 +31,17 @@ public class NightscoutUploader: NSObject {
...
+    var treatmentsQueue: [NightscoutTreatment]
     
     var lastMeterMessageRxTime: NSDate?
     
-    var observingPumpEventsSince: NSDate
+    public private(set) var observingPumpEventsSince: NSDate
+    
+    var lastStoredTreatmentTimestamp: NSDate = NSDate.distantPast() {
+        didSet {
+            NSUserDefaults.standardUserDefaults().lastStoredTreatmentTimestamp = lastStoredTreatmentTimestamp
```
loudnate, line 42: This is problematic if the user changes their pump ID, no?

loudnate:
So I also plan to solve this problem in LoopKit's DoseStore, which will have a delegate protocol to trigger uploading records from the SQLite database, and maintains an `uploaded` flag. 

We can revisit when we get that far.

NSUserDefaults.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  NSUserDefaults.swift
+//  RileyLink
```
loudnate: This file should be in the NightscoutUploadKit framework

ps2: Thanks, good catch.

loudnate:
LGTM

PR #132, opened 2016-06-24
Fix precision of reservoir value

ps2:
Nightscout was showing many more digits of precision than desired, due to float representation rounding.  Maybe this is nightscout's responsibility, but it's easier for me to tackle it here.

### DIFF
--- a/NightscoutUploadKit/NightscoutUploader.swift
+++ b/NightscoutUploadKit/NightscoutUploader.swift
@@ -169,7 +169,7 @@ public class NightscoutUploader: NSObject {
                 "timestamp": pumpDateStr,
                 "bolusiob": status.iob,
             ],
-            "reservoir": status.reservoirRemainingUnits,
+            "reservoir": Double(round(status.reservoirRemainingUnits*1000)/1000,
             "battery": [
                 "percent": status.batteryRemainingPercent
             ]


loudnate:
Isn't JavaScript awesome?

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -169,7 +169,7 @@ public class NightscoutUploader: NSObject {
                 "timestamp": pumpDateStr,
                 "bolusiob": status.iob,
             ],
-            "reservoir": status.reservoirRemainingUnits,
+            "reservoir": Double(round(status.reservoirRemainingUnits*1000)/1000,
```
loudnate, line 172: How does this change the output JSON?

ps2, line 172: I was a bit suspicious of this solution, too, but I did a test, and it appears to work. I think it's actually a rounding error from earlier math, rather than a serialization issue, so rounding here works. If this fails, it's easy enough to back out.

loudnate, line 172: Wait, a rounding error in the decoded message? `status.reservoirRemainingUnits`?? We'd need to fix that!

I also think I would have seen it by now...

ps2, line 172: Well, it manifests like this: 0.07500000000000001 instead of 0.075.  Not a serious issue in terms of accuracy.

ps2, line 172: That last value was an example from bolusiob, which I just noticed has the same issue.   Here is an example value from the reservoir field: 42.90000000000001

ps2:
You explain it to me. :)

```
        let testDouble = Double(Int(429)) * 0.1
        let testDouble2 = Double(round(testDouble*1000)/1000)
        print("testDouble = \(testDouble)")
        print("testDouble2 = \(testDouble2)")
        let jsonData = [
            "testDouble": testDouble,
            "testDouble2": testDouble2,

        ]
        do {
            let sendData = try NSJSONSerialization.dataWithJSONObject(jsonData, options: NSJSONWritingOptions.PrettyPrinted)
            let sendDataStr = String(data: sendData, encoding: NSUTF8StringEncoding)
            print("sendDataStr = \(String(sendDataStr))")
        } catch {
            print("Could not serialize json")
        }
```

Output:

```
testDouble = 42.9
testDouble2 = 42.9
sendDataStr = Optional("{\n  \"testDouble\" : 42.90000000000001,\n  \"testDouble2\" : 42.9\n}")
```

NightscoutUploadKit/NightscoutUploader.swift
loudnate, line 172: Interesting, well let's change the data type to Double in the message then, which should help. 

Reservoir values are only given to the 0.1, though, which shouldn't have precision issues with a float32...

loudnate:
Try it again, dividing by 10 instead. :)

loudnate:
Changing the first line:

`let testDouble = Double(Int(429)) / 10`

outputs like this:

```
testDouble = 42.9
testDouble2 = 42.9
sendDataStr = Optional("{\n  \"testDouble\" : 42.9,\n  \"testDouble2\" : 42.9\n}")
```

loudnate:
I'm fine with merging this, but I also proposed #133 to get rid of the Decimal \* Decimal cases.

ps2:
Abandoning in favor of #133

PR #133, opened 2016-06-25
Fixing floating-point precision in derived message values

loudnate:


ps2:
I'm fine with this over #132. Only the compiler gods will know why this is necessary.

loudnate:
Yeah, unfortunately. The #132 is essentially doing the same thing, relying on float division rather than multiplication.

http://stackoverflow.com/a/4125154

PR #135, opened 2016-06-25
Rl rename

ps2:
Reinstate RileyLink renaming feature.  There is a delay after renaming, before we get the iOS callbacks that indicate the device is renamed.  There should maybe be some progress UI to show that the renaming is happening, so users don't keep trying to rename.  But going with this for now.


ps2:
Wrong branch

PR #136, opened 2016-06-25
Rl rename

ps2:
Reinstate RileyLink renaming feature.  There is a delay after renaming, before we get the iOS callbacks that indicate the device is renamed.  There should maybe be some progress UI to show that the renaming is happening, so users don't keep trying to rename.  But going with this for now.


RileyLinkBLEKit/RileyLinkBLEDevice.m
```diff
@@ -426,10 +430,6 @@ - (void)peripheral:(CBPeripheral *)peripheral didUpdateNotificationStateForChara
     NSLog(@"Updated notification state for %@, %@", characteristic, error);
 }
 
-- (void)peripheralDidUpdateName:(CBPeripheral *)peripheral {
-    [[NSNotificationCenter defaultCenter] postNotificationName:RILEYLINK_EVENT_LIST_UPDATED object:nil];
```
loudnate, line 430: Are we no longer using this notification name? Can we remove it?

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -184,6 +184,8 @@ class PumpOpsSynchronous {
                 
                 let response: ReadTempBasalCarelinkMessageBody = try getMessageBodyWithType(.ReadTempBasal)
                 
+                print("temp basal results: \(duration) \(response.rateType)")
```
loudnate: Left over from debugging?

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -131,6 +159,9 @@ public class RileyLinkDeviceTableViewController: UITableViewController {
         switch Section(rawValue: indexPath.section)! {
         case .Device:
             switch DeviceRow(rawValue: indexPath.row)! {
+            case .CustomName:
+                cell.textLabel?.text = NSLocalizedString("Name", comment: "The title of the cell showing device name")
+                cell.detailTextLabel?.text = device.name
```
loudnate, line 164: `cell.accessoryType = .disclosureIndicator`?

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -379,14 +432,22 @@ public class RileyLinkDeviceTableViewController: UITableViewController {
...
                 vc = CommandResponseViewController { [unowned self] (completionHandler) -> String in
-                    self.device.ops?.pressButton({ (response) in
-                        dispatch_async(dispatch_get_main_queue()) {
-                            switch response {
-                            case .Success(let msg):
-                                completionHandler(responseText: "Result: " + msg)
-                            case .Failure(let error):
-                                completionHandler(responseText: String(error))
-                            }
+//                    self.device.ops?.pressButton({ (response) in
```
loudnate: More debugging code?

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -414,8 +475,34 @@ public class RileyLinkDeviceTableViewController: UITableViewController {
...
+                    device.setCustomName(controller.value!)
+                default:
+                    break
+                }
+            default:
+                break
+
+            }
+
+            tableView.reloadData()
```
loudnate: No need to broadly reload here, since you're already updating the display after the notification comes in.

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
ps2: Yikes!  Yeah, I wasn't seeing any temp basals being enacted in Loop, so I wanted to try in RL, so I could see what was going on.  Turned out I had increased my max basal in Loop, but not on the pump, so it was failing. Thanks so much for catching this!

RileyLinkBLEKit/RileyLinkBLEDevice.m
ps2, line 430: Still using it here: https://github.com/ps2/rileylink_ios/blob/master/RileyLinkKit/RileyLinkDeviceManager.swift#L37

loudnate:
LGTM!

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
loudnate: ah, sorry about that. settings sync is something I hope to consider tackling soon.

PR #137, opened 2016-06-26
{WIP} Worlwide pump support

oskarpearson:
Initial code to better support worldwide pumps. Unfortunately it's not quite working yet

Feedback welcome. One option discussed in Gitter chat is to auto-detect the frequency, then keep state. I think that is the right long term solution, but this seemed easier :)

### DIFF
--- a/RileyLink/Config.h
+++ b/RileyLink/Config.h
@@ -20,5 +20,6 @@
 @property (nonatomic, nullable, strong) NSTimeZone *pumpTimeZone;
 @property (nonatomic, nullable, strong) NSSet *autoConnectIds;
 @property (nonatomic, assign) BOOL uploadEnabled;
+@property (nonatomic, assign) BOOL worldwideRadioLocale;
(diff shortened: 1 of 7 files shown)


oskarpearson:
@loudnate You mentioned you wanted to comment. Beware - first ever swift code: I have no idea what I'm doing :)

ps2:
Could you re-submit this against dev?

loudnate:
I think direction-wise there's a few things we should do here. First, let's get the frequencies out of the mutable arrays on PumpState and into the Ops call.

Initially we can try both ranges, and if we need to optimize, store a "US/WW/Unknown" flag on PumpState once we have a match.

oskarpearson:
Hi there.

I've been a little busy since the referendum - but hope to sort this out in the next few days. I'll refactor as soon as I get a moment.

Thanks for the suggestions - I think they make sense.

Oskar

PR #149, opened 2016-07-11
Fix parsing of date fields in mysentry pump status records

ps2:
And, thanks to @beached for figuring this out: parse new clock type field to denote desired display of 24 hour or 12 hour clock.


ps2:
wrong target branch.

PR #150, opened 2016-07-11
Merging accidental squash commit back to dev.

loudnate:


loudnate:
actually I don't think we want to do this – I think it's best we re-merge the actual commits from dev to master.

PR #155, opened 2016-07-15
Add journal entry meal marker pump event

ps2:


MinimedKit/PumpEvents/JournalEntryMealMarkerPumpEvent.swift
```diff
@@ -0,0 +1,33 @@
...
+//  JournalEntryMealMarkerPumpEvent.swift
+//  RileyLink
+//
+//  Created by Pete Schwamb on 7/14/16.
+//  Copyright © 2016 Pete Schwamb. All rights reserved.
+//
+
+import Foundation
+
+public struct JournalEntryMealMarkerPumpEvent: TimestampedPumpEvent {
```
loudnate, line 11: I added this to decocare a while back; I used to use this feature a lot on my pump 😄

There's a carb value here too: https://github.com/bewest/decoding-carelink/blob/313b13b6f465c91d5b01fdaec2b6c4327cabda5f/decocare/history.py#L241

loudnate:
LGTM. You can add `.JournalEntryOtherMarker = 0x43` too with just one line in `PumpEventType.swift` – 7-byte messages will now use the default struct.

ps2:
@loudnate I wonder if it makes sense to have a HKQuantity type object for tracking this kind of thing. It's probably too easy right now to do event.carbohydrates, and assume grams.  But I don't want to do redundant carbGrams and carbExchange entries either.

loudnate:
@ps2 the only pause I'd have is the dependency on HealthKit; otherwise I completely love `HKQuantity`... I'm so glad they're extending the concept to [Foundation](https://developer.apple.com/reference/foundation/measurement) with `NSMeasurement`, which might be my favorite feature about the iOS 10 SDK.

loudnate:
Actually `HKQuantity` is available to all devices, (e.g. iPad) regardless of whether it supports `HKHealthStore`. It would vastly improve references to carbs and glucose.

PR #156, opened 2016-07-16
Extract another high bit to accept bgcal entries up to 600

ps2:
Commit message has a typo. The pump doesn't actually go above 600.


loudnate:
LGTM

PR #157, opened 2016-07-16
JournalEntryOther

ps2:
Allow use of placeholder enum types for 7 byte pump events, while still keeping a useful event type name in the dictionary rep.

### DIFF
--- a/MinimedKit/PumpEvents/PlaceholderPumpEvent.swift
+++ b/MinimedKit/PumpEvents/PlaceholderPumpEvent.swift
@@ -13,7 +13,6 @@ public struct PlaceholderPumpEvent: TimestampedPumpEvent {
     public let length: Int
     public let rawData: NSData
     public let timestamp: NSDateComponents
-    public let name: String
 
     public init?(availableData: NSData, pumpModel: PumpModel) {
         length = 7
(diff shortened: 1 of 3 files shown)


MinimedKit/PumpEvents/PlaceholderPumpEvent.swift
```diff
@@ -9,26 +9,33 @@
 import Foundation
 
 
-public struct UnknownPumpEvent: TimestampedPumpEvent {
+public struct PlaceholderPumpEvent: TimestampedPumpEvent {
     public let length: Int
     public let rawData: NSData
     public let timestamp: NSDateComponents
+    public let name: String
```
loudnate: Rather than storing the string in the struct (which requires heap allocation), let's just move the name generation code to the `dictionaryRepresentation` impl.

MinimedKit/PumpEvents/PlaceholderPumpEvent.swift
```diff
@@ -9,26 +9,33 @@
...
+            name = String(type).componentsSeparatedByString(".").last!
+        } else {
+            name = "UnknownPumpEvent(\(rawData[0] as UInt8))"
+        }
     }
 
     public var dictionaryRepresentation: [String: AnyObject] {
         return [
-            "_type": "UnknownPumpEvent(\(rawData[0] as UInt8))",
+            "_type": "\(name)",
```
loudnate: simplify to `"_type": name,`

MinimedKit/PumpEvents/PlaceholderPumpEvent.swift
ps2: Hah. I find this amusing, because it was my previous style (in minimed_rf) to calculate everything as needed from the data, but then I saw you doing calc on init, and decided I liked that better as well.  Guessing you have a reason for why you consider this different, but still... :)

loudnate:
LGTM

MinimedKit/PumpEvents/PlaceholderPumpEvent.swift
loudnate: Ha! Well feel free to not listen to me. This case felt unique because it was storing the Type name as a string in itself. 

Now that we're keeping the raw bytes around, we could move them to computed properties, but we'd also have to keep the PumpModel around too.

PR #160, opened 2016-07-17
Insulin marker

ps2:


MinimedKit/PumpEvents/JournalEntryInsulinMarkerPumpEvent.swift
```diff
@@ -0,0 +1,39 @@
...
+            return nil
+        }
+        
+        rawData = availableData[0..<length]
+        
+        timestamp = NSDateComponents(pumpEventData: availableData, offset: 2)
+        
+        let lowBits = rawData[1] as UInt8
+        let highBits = rawData[4] as UInt8
+        amount = Double((Int(highBits & 0b1100000) << 3) + Int(lowBits)) / 10.0
```
loudnate, line 30: I think there's a convenience initializer for Int called `bigEndianBytes` that you can use here (and other places) to simplify

loudnate:
LGTM

MinimedKit/PumpEvents/JournalEntryInsulinMarkerPumpEvent.swift
ps2, line 30: No, this isn't contiguous.  (byte 1 has low bits and byte 4 has high)

loudnate, line 30: Right, I assumed you'd apply the mask and combine into an array, but I'm not sure how much code that actually saves :)

Issue #147, opened 2016-07-04
Add a method to test/verify Nightscout credentials


loudnate:
Can we add a method that does a no-op call to the Nightscout API using the provided credentials, to help users validate that it's working immediately after entering them?

@ps2 as someone who still hasn't setup Nightscout, I'm looking to you for guidance. The API endpoints I've seen seem to be read-only public; is that true for everybody? I'm hoping there's a way test both the URL and the secret.

ps2:
@mhaeberli mentioned an endpoint specifically for testing auth. Something like `https://xxxsite.azurewebsites.net/api/v1/experiments/test`  

Just adding as a note for when I look into this.

loudnate:
https://github.com/nightscout/cgm-remote-monitor/wiki/API-v1.0.0-beta-Security

PR #161, opened 2016-07-19
Validate ns auth params

ps2:
Adds a method to check Nightscout authentication parameters.

Incorporates ServiceAuthentication models and views from Loop for use in settings.

Also there is an API level compatibility breaking change: NightscoutUploader now takes an actual NSURL instead of a string.


ps2:
Fixes #147

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -57,7 +58,7 @@ public class NightscoutUploader {
         lastStoredTreatmentTimestamp = nil
     }
 
-    public init(siteURL: String?, APISecret: String?) {
+    public init(siteURL: NSURL?, APISecret: String?) {
```
loudnate: Now that this is wrapped in the `NightscoutService` struct, which only creates it when these two arguments are non-nil, you could simplify here and make these non-optional

RileyLink/KeychainManager+RileyLink.swift
```diff
@@ -0,0 +1,40 @@
...
+
+import Foundation
+
+
+private let NightscoutAccount = "NightscoutAPI"
+
+
+extension KeychainManager {
+    
+    func setNightscoutURL(URL: NSURL?, secret: String?) throws {
```
loudnate: This might be a bug in the Loop code – but nothing actually throws here.

RileyLink/Models/NightscoutService.swift
```diff
@@ -0,0 +1,87 @@
...
+        
+        guard let siteURL = siteURL, APISecret = APISecret else {
+            completion(success: false, error: nil)
+            return
+        }
+        self.uploader = NightscoutUploader(siteURL: siteURL, APISecret: APISecret)
+        
+        self.uploader?.checkAuth({ (error) in
+            if let error = error {
+                completion(success: false, error: error)
```
loudnate, line 74: Line 72.5 should be `self.isAuthorized = false`

loudnate, line 74: It's a bit of an indeterminate spot in the this `ServiceAuthentication` state machine, unfortunately; `verify` sets the `isAuthorized` state when it knows it, but doesn't require `isAuthorized` to be true to have the credentials set. It's really just a warning to the user.

loudnate:
LGTM!

PR #168, opened 2016-07-21
Fix crash with timer being doubly invalidated

ps2:


RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -15,7 +15,7 @@ public class RileyLinkDeviceTableViewController: UITableViewController, TextFiel
 
     public var device: RileyLinkDevice!
     
-    var rssiFetchTimer: NSTimer!
```
loudnate, line 18: Every time you write an exclamation mark, someone at Apple deletes a Radar. 😉

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -82,14 +80,18 @@ public class RileyLinkDeviceTableViewController: UITableViewController, TextFiel
         if appeared {
             tableView.reloadData()
         }
+        
+        rssiFetchTimer = NSTimer.scheduledTimerWithTimeInterval(3, target: self, selector: #selector(updateRSSI), userInfo: nil, repeats: true)
```
loudnate, line 88: `viewWillAppear` can run multiple times before `viewDidDisappear` runs, for example when cancelling a swift-from-left-edge gesture after starting it.

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
ps2, line 18: Yep.  Feel like I just got my hand smacked for reaching for the cookie jar. :)

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -15,7 +15,7 @@ public class RileyLinkDeviceTableViewController: UITableViewController, TextFiel
 
     public var device: RileyLinkDevice!
     
-    var rssiFetchTimer: NSTimer!
+    var rssiFetchTimer: NSTimer?
```
loudnate: Use `didSet` to invalidate, to make sure you're always calling `invalidate` before replacing the reference.

ps2: great idea.

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
loudnate, line 88: But `viewWillDisappear` should be balanced with `viewWillAppear`, so let's set `rssiFetchTimer` to `nil` in that method.

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -82,14 +80,18 @@ public class RileyLinkDeviceTableViewController: UITableViewController, TextFiel
...
+        rssiFetchTimer = NSTimer.scheduledTimerWithTimeInterval(3, target: self, selector: #selector(updateRSSI), userInfo: nil, repeats: true)
 
         appeared = true
     }
     
     public override func viewDidDisappear(animated: Bool) {
         super.viewDidDisappear(animated)
         
-        rssiFetchTimer.invalidate()
+        if let timer = rssiFetchTimer {
```
loudnate: with the `didSet` change above, you can delete these three lines.

loudnate:
I'm sure before long you'll hate `NSTimer` as much as I've told you I do 😄 

Here's a reference of how I use it, clunkily: https://github.com/loudnate/Loop/blob/master/Loop/Views/LoopCompletionHUDView.swift

loudnate:
There is one more edge case where the app becomes inactive while the view controller is visible. I think it's fine to ignore for now.

loudnate:
LGTM

loudnate:
Can you send a PR to Loop with the updated frameworks after you release this? I'm away from a computer today.

ps2:
sure

PR #175, opened 2016-07-28
Device status improvements

ps2:
New swift structures to represent Nightscout device status upload api json.  It's infrastructure for uploading Loop status to NS, and also for uploading phone status separately from pump status.


NightscoutUploadKit/DeviceStatus/BatteryStatus.swift
```diff
@@ -0,0 +1,35 @@
...
+//  BatteryStatus.swift
+//  RileyLink
+//
+//  Created by Pete Schwamb on 7/28/16.
+//  Copyright © 2016 Pete Schwamb. All rights reserved.
+//
+
+import Foundation
+
+public class BatteryStatus {
```
loudnate: Any reason to use classes rather than structs?

loudnate:
LGTM

NightscoutUploadKit/DeviceStatus/BatteryStatus.swift
ps2: Just that I haven't been indoctrinated fully.  Getting closer, thanks to you. :)

NightscoutUploadKit/DeviceStatus/IOBStatus.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+//  Created by Pete Schwamb on 7/28/16.
+//  Copyright © 2016 Pete Schwamb. All rights reserved.
+//
+
+import Foundation
+
+public struct IOBStatus {
+    let iob: Double // basal iob + bolus iob: can be negative
+    let basaliob: Double
```
loudnate, line 13: `basalIOB` would be more cocoa-like.

PR #177, opened 2016-07-31
Adding a stateless upload path for Nightscout treatments

loudnate:


ps2:
This looks good. I'm curious as to how you'll handle upload failures in loop, but I guess that's my next place to review, and this change seems useful independent of what you're actually doing in loop.

loudnate:
It will have a similar retry frequency (next history add, e.g. 5 min) as you have today. Changing this to `NSURLSessionUploadTask` is also a hint to the system to allow the upload to run in the background.

I ran the stack overnight (with simulated uploading) and everything went well. History data and its upload status is also visible in the LoopKit-provided view controllers.

PR #178, opened 2016-08-01
Bundle up fetching of common pump status items into single call

ps2:
New command for fetching of bolusing and suspended states. 
Replaced "read remaining insulin" call with "read pump status"

### DIFF
--- a/RileyLinkKit/PumpOps.swift
+++ b/RileyLinkKit/PumpOps.swift
@@ -156,8 +156,8 @@ public class PumpOps {
      This operation is performed asynchronously and the completion will be executed on an arbitrary background queue.
 
      - parameter completion: A closure called after the command is complete. This closure takes a single Result argument:
-     - Success(status): A structure describing the current status of the pump
-     - Failure(error): An error describing why the command failed
+        - Success(status): A structure describing the current status of the pump
+        - Failure(error): An error describing why the command failed
      */
     public func readPumpStatus(completion: (Either<PumpStatus, ErrorType>) -> Void) {
         device.runSession { (session) in
--- a/RileyLinkKit/PumpOps.swift
+++ b/RileyLinkKit/PumpOps.swift
@@ -149,6 +149,30 @@ public class PumpOps {
         }
     }
 
+
+    /**
+     Reads clock, reservoir, battery, bolusing, and suspended state from pump
+
+     This operation is performed asynchronously and the completion will be executed on an arbitrary background queue.
+
+     - parameter completion: A closure called after the command is complete. This closure takes a single Result argument:
+     - Success(status): A structure describing the current status of the pump
+     - Failure(error): An error describing why the command failed
+     */
+    public func readPumpStatus(completion: (Either<PumpStatus, ErrorType>) -> Void) {
+        device.runSession { (session) in
+            let ops = PumpOpsSynchronous(pumpState: self.pumpState, session: session)
+
+            do {
+                let response: PumpStatus = try ops.readPumpStatus()
+                completion(.Success(response))
+            } catch let error {
+                completion(.Failure(error))
+            }
+        }
+    }
+
+
     /**
      Sets a bolus
      
--- a/RileyLinkKit/PumpOpsSynchronous.swift
+++ b/RileyLinkKit/PumpOpsSynchronous.swift
@@ -411,6 +411,41 @@ class PumpOpsSynchronous {
         }
         return frameData
     }
+
+    internal func readPumpStatus() throws -> PumpStatus {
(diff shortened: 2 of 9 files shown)


ps2:
This is mainly to support x22 pumps status reporting in NS

RileyLinkKit/PumpOps.swift
```diff
@@ -149,6 +149,30 @@ public class PumpOps {
...
     }
 
+
+    /**
+     Reads clock, reservoir, battery, bolusing, and suspended state from pump
+
+     This operation is performed asynchronously and the completion will be executed on an arbitrary background queue.
+
+     - parameter completion: A closure called after the command is complete. This closure takes a single Result argument:
+     - Success(status): A structure describing the current status of the pump
```
loudnate: nit, but this line and the one below it should be tabbed over so Apple's markdown doc generation shows them as bullet points.

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -411,6 +411,41 @@ class PumpOpsSynchronous {
...
+
+public struct PumpStatus {
+    let clock: NSDateComponents
+    let batteryVolts: Double
+    let suspended: Bool
+    let bolusing: Bool
+    let reservoir: Double
+    let model: PumpModel
+
+    public init(clock: NSDateComponents, batteryVolts: Double, suspended: Bool, bolusing: Bool, reservoir: Double, model: PumpModel) {
```
loudnate: No need to make this initializer public, right? If so you can delete it entirely since an (internal) initializer is automatically generated for structs.

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -477,21 +477,25 @@ public class RileyLinkDeviceTableViewController: UITableViewController, TextFiel
...
                         dispatch_async(dispatch_get_main_queue()) {
                             switch result {
-                            case .Success(let units):
-                                completionHandler(responseText: String(format: NSLocalizedString("%1$@ Units remaining", comment: "The format string describing units of insulin remaining: (1: number of units)"), self.decimalFormatter.stringFromNumber(units)!))
+                            case .Success(let status):
+                                var str = String(format: NSLocalizedString("%1$@ Units of insulin remaining\n", comment: "The format string describing units of insulin remaining: (1: number of units)"), self.decimalFormatter.stringFromNumber(status.reservoir)!)
+                                str += String(format: NSLocalizedString("Battery: %1$@ volts\n", comment: "The format string describing pump battery voltage: (1: battery voltage)"), self.decimalFormatter.stringFromNumber(status.batteryVolts)!)
+                                str += String(format: NSLocalizedString("Suspended: %1$@\n", comment: "The format string describing pump suspended state: (1: suspended)"), status.suspended.description)
```
loudnate: Per the docs:

> String(instance) will work for an instance of any type, returning its description if the instance happens to be CustomStringConvertible. Using CustomStringConvertible as a generic constraint, or accessing a conforming type's description directly, is therefore discouraged.

`bolusing.description -> String(bolusing)`

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -411,6 +411,41 @@ class PumpOpsSynchronous {
...
+
+        let resResp: ReadRemainingInsulinMessageBody = try getMessageBodyWithType(.ReadRemainingInsulin)
+
+        let reservoir = resResp.getUnitsRemainingForStrokes(pumpModel.strokesPerUnit)
+
+        let battResp: GetBatteryCarelinkMessageBody = try getMessageBodyWithType(.GetBattery)
+
+        let statusResp: ReadPumpStatusMessageBody = try getMessageBodyWithType(.ReadPumpStatus)
+
+        return PumpStatus(clock: clockResp.dateComponents, batteryVolts: battResp.volts, suspended: statusResp.suspended, bolusing: statusResp.bolusing, reservoir: reservoir, model: pumpModel)
```
loudnate: Did you say the battery status had the quarter-percentage encoded as well? Should we return that too?

RileyLinkKit/PumpOpsSynchronous.swift
ps2: cool.  Didn't know that.

RileyLinkKit/PumpOpsSynchronous.swift
ps2: I don't know of a quarter-percentage encoded value outside of mysentry.   There _is_ a status flag (low/normal) that we're not decoding yet.

PR #179, opened 2016-08-02
update RileyLink app version to reflect updated since master 7.1

mhaeberli:
(as …7.2d)

propose updating dev branch info string to reflect it's more recent than production 7.1 and it's dev branch.

### DIFF
--- a/RileyLink/RileyLink-Info.plist
+++ b/RileyLink/RileyLink-Info.plist
@@ -19,7 +19,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>0.7.1</string>
+	<string>0.7.2d</string>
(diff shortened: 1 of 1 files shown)


loudnate:
Version strings are updated using agvtool, generally with `agvtool bump` and `agvtool new-marketing-version 0.7.1`. 

It's hard to pre-select a marketing version without knowing the compatibility of the changes it will include. 

Any reason why you're running the dev branch in particular?

mhaeberli:
Nathan,
no - just thought we'd give it a try.
but I see your point.
I just imagined that it would be good to distinguish it as _later_ than
0.7.1 ...
I am very happy that we have version numbers now, in Loop and RileyLink !
Thanks,
Martin

ps2:
Thanks @mhaeberli.  Yes, we typically version dev when we're ready to cut a release.

PR #180, opened 2016-08-02
X22 device status

ps2:


RileyLink/DeviceDataManager.swift
```diff
@@ -242,46 +370,45 @@ class DeviceDataManager {
 
     init() {
```
loudnate, line 371: Sorry to burden you with the haphazard expansion of Loop's DDM. I promise this will be refactored as things settle.

ps2, line 371: No worries; I'm just bringing RL up to speed with Loop, so when you do the refactoring, I can ~~steal~~ reuse your awesome ideas more easily. ;)

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -435,19 +435,20 @@ class PumpOpsSynchronous {
...
 
 public struct PumpStatus {
-    let clock: NSDateComponents
-    let batteryVolts: Double
-    let batteryStatus: BatteryStatus
-    let suspended: Bool
-    let bolusing: Bool
-    let reservoir: Double
-    let model: PumpModel
+    public let clock: NSDateComponents
```
loudnate, line 444: ah sorry for not catching this sooner!

loudnate:
LGTM ++

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -122,20 +122,6 @@ public class NightscoutUploader {
         uploadToNS(treatments, endpoint: defaultNightscoutTreatmentPath, completion: completionHandler)
     }
 
-    public func getPumpStatusFromMySentryPumpStatus(status: MySentryPumpStatusMessageBody) throws -> PumpStatus {
```
loudnate, line 125: Why did you remove this, exactly? Is there an alternative method for uploading these values from MySentry records, or were you thinking you wanted all pumps to wake+poll (which I, personally, would not want to do)

ps2, line 125: That was just the logic to convert MySentryPumpStatusMessageBody into PumpStatus.  I moved it into DDM, as it felt out of place in the uploader (really uses no state of the uploader).  https://github.com/ps2/rileylink_ios/blob/dev/RileyLink/DeviceDataManager.swift#L212-L226

Looking at that, I see I forgot to convert the throw.

Where would you suggest logic like this resides?  While I don't want to clutter up DDM too much, things that bridge frameworks (like this method bridges MinimedKit and NightscoutUploadKit models) seem to be ending up here (in DDM) at the high level, and that at least does have the advantage of keeping the frameworks focused.

ps2, line 125: And yeah, I have no intention of not uploading from MySentry or forcing polling.

loudnate, line 125: If you're OK with MinimedKit remaining a dependency on Nightscout upload it, then I'd make it a convenience initializer on PumpStatus. 

I didn't see it moved to DDM, sorry - I think that's a perfectly reasonable step in a path to allow NightscoutUloadKit to stand alone

PR #183, opened 2016-08-03
Loop optional values, and COB status

ps2:
Make certain values optional for NS upload, as we might not have them. Also, add carbs on board status for reporting, since Loop uses carbs in its calculations.


loudnate:
LGTM

PR #187, opened 2016-08-04
Timer tick always on

ps2:
Ran this overnight and today, and it seems to be working well.


RileyLink/DeviceDataManager.swift
```diff
@@ -301,8 +292,13 @@ class DeviceDataManager {
                 }
             })
         }
+
+        if lastHistoryAttempt == nil || lastHistoryAttempt!.timeIntervalSinceNow < (-5 * 60) {
```
loudnate: `(-5 * 60)` can be swapped with `NSTimeInterval(minutes: -5)`

loudnate:
LGTM

PR #190, opened 2016-08-08
Upload predicted bgs

ps2:
There are two different places predicted BGs can go; recommending we only use the one on LoopSuggested, but keeping both in case we want to use that feature.

### DIFF
--- a/NightscoutUploadKit/DeviceStatus/LoopEnacted.swift
+++ b/NightscoutUploadKit/DeviceStatus/LoopEnacted.swift
@@ -13,14 +13,14 @@ public struct LoopEnacted {
     let duration: NSTimeInterval
     let timestamp: NSDate
     let received: Bool
-    let predictedBGValues: [Double]?
+    let predBGs: [Double]?
     
-    public init(rate: Double, duration: NSTimeInterval, timestamp: NSDate, received: Bool, predictedBGValues: [Double]?) {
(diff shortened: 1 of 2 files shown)


NightscoutUploadKit/DeviceStatus/LoopEnacted.swift
```diff
@@ -13,20 +13,28 @@ public struct LoopEnacted {
     let duration: NSTimeInterval
     let timestamp: NSDate
     let received: Bool
+    let predBGs: [Double]?
     
-    public init(rate: Double, duration: NSTimeInterval, timestamp: NSDate, received: Bool) {
+    public init(rate: Double, duration: NSTimeInterval, timestamp: NSDate, received: Bool, predBGs: [Double]? = nil) {
```
loudnate, line 18: How are units managed?

ps2, line 18: BG units in NS are always stored in mg/dl, and are converted at display time, ~~if an environment variable is set~~.

ps2, line 18: Actually, setting the env var DISPLAY_UNITS didn't seem to do anything.  There is a preference on the settings ui that can switch between mmol and mg/dl, and that does seem to work.

loudnate:
LGTM

PR #192, opened 2016-08-09
Creating convenience initializers for Nightscout Status structs

loudnate:
Using more natural arguments for the device/system, rather than surfacing NS design decisions any higher than necessary


ps2:
Thanks, these look great!

loudnate:
Feel free to merge as needed,

PR #193, opened 2016-08-09
Carb and insulin effect points

ps2:
Going to need a little updating to work with the HK initializers.


NightscoutUploadKit/DeviceStatus/PredictedBG.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+
+import Foundation
+
+public struct PredictedBG {
+    let values: [Double]
+    let cob: [Double]?
+    let iob: [Double]?
+
+    public init(values: [Double], cob: [Double]? = nil, iob: [Double]? = nil) {
```
loudnate: How do you know what dates these values are associated with?

ps2: NS uses the timestamp on the containing structure as the start date, and assumes a 5 minute interval thereafter. I could add a timestamp field to this, and we can later update NS to use it.  Do you think one base timestamp for all points, or separate base timestamps, one for each array (values, cob, iob)?  Seems like the former would suffice.

loudnate: ugh. well it sounds like the NS start date should be exactly the date of the first prediction value then...

Why does NS need the trailing cob/iob, starting from time = now? it's not all that useful information.

ps2: NS doesn't _need_ them; they're optional.  It's used to display a glucodyn like graph, where the carb effect line is shown on top, insulin on bottom, and combined (actual bg): https://files.gitter.im/nightscout/public/als3/Screenshot-2016-06-03-23.37.04.png

For me, it's a visual way of explaining why we think the bg will go where we think it will.

loudnate: Interesting, so it should upload carb/insulin effect, not cob/iob, then?

ps2: For the start date of the points, the timestamp on the parent structure (LoopSuggested, or LoopEnacted) is used.  I'm thinking that for LoopSuggested, it's going to usually match? i.e. the start time of the forecasted points is going to match recommendedDate on TempBasalRecommendation?  If not, I can add a timestamp that is the actual start time of the forecasted points to this structure, and submit a patch to NS to use it.

ps2: Yes, RE effect vs *ob; that's just the (misleading) name on the structs.  They are actually predicted bg + effect values.

loudnate: ok, the PR open in Loop was fetching IOB/COB.

ps2: That was for current IOB/COB; I haven't added the forecasted values to the PR yet, as I just realized what they were, today, while talking to @jasoncalabrese

loudnate: gotcha. I didn't see COB used anywhere, but `CarbStore` and `DoseStore` both have interfaces for fetching a single value.

NightscoutUploadKit/DeviceStatus/LoopSuggested.swift
```diff
@@ -17,10 +17,9 @@ public struct LoopSuggested {
     let reason: String?
     let tick: Int?
     let correction: Double?
-    let predBGs: [Double]?
+    let predBGs: PredictedBG?
 
-
-    public init(timestamp: NSDate, rate: Double, duration: NSTimeInterval, eventualBG: Int, bg: Int, reason: String? = nil, tick: Int? = nil, correction: Double? = nil, predBGs: [Double]? = nil) {
+    public init(timestamp: NSDate, rate: Double, duration: NSTimeInterval, eventualBG: Int, bg: Int, reason: String? = nil, tick: Int? = nil, correction: Double? = nil, predBGs: PredictedBG? = nil) {
```
loudnate: `HKQuantity` for `bg` and `eventualBG`?

ps2: What do you think about just having HKQuantity initializers, and dropping the double based ones?

loudnate: That'd be great. I wasn't sure if you felt like it's important to maintain a "pure" initializer for the struct.

ps2:
Showing too many commits in the diff that are already on dev, so going to recreate.

PR #194, opened 2016-08-09
Carb and insulin effect points

ps2:
Allows us to upload carb and insulin effect points to NS.  Also changes initializer signatures to use HKQuantity for bg values.


loudnate:
LGTM

PR #197, opened 2016-08-11
v0.9.0

ps2:
- Compatibility breakage: Nightscout upload structures and initializers
- Keep RileyLink BLE timer tick on, to prevent app from not waking.
- Remove unused files that were causing many compilation warnings.
- Updates to support uploading Loop status to Nightscout, specifically predicted BGs
- History fetching more reliable for x22 pumps.


loudnate:
Whoo! LGTM.

PR #199, opened 2016-08-17
Adding EnableBolusWizardPumpEvent

ps2:


MinimedKit/PumpEvents/EnableBolusWizardPumpEvent.swift
```diff
@@ -0,0 +1,33 @@
...
+//  EnableBolusWizardPumpEvent.swift
+//  RileyLink
+//
+//  Created by Pete Schwamb on 8/16/16.
+//  Copyright © 2016 Pete Schwamb. All rights reserved.
+//
+
+import Foundation
+
+public struct EnableBolusWizardPumpEvent: TimestampedPumpEvent {
```
loudnate, line 11: I think we ended up with a default event for these empty-body events, right? They print the case name for the event type.

That said, there's probably a bit somewhere that indicates whether the bolus wizard was enabled or disabled...

loudnate:
LGTM

PR #201, opened 2016-08-19
Ns loop plugin

ps2:
Populate data for Nightscout loop plugin. https://github.com/ps2/cgm-remote-monitor/tree/loop-plugin


loudnate:
LGTM

PR #203, opened 2016-08-21
Handle timestamp offset for bolus wizard usage.

ps2:
As reported by @elnjensen, sometimes bolus wizard usages are being reported as plain boluses.  From the history dumps we saw that this happens when the bolus and bolus wizard events are offset by a second.

This patch handles that case by allowing up to a 2 second difference between the bolus wizard and bolus event.


loudnate:
LGTM

PR #207, opened 2016-08-27
Customizing the conformance of some PumpEvents to CustomStringConvertible

loudnate:
This is great place for contributors to step in, especially if they want to change the string representations they see for history events.


ps2:
LGTM!

PR #211, opened 2016-08-29
History time fix

ps2:
Avoid fetching history events that don't fall on expected timeline, as that indicates a non-recoverable clock change

### DIFF
--- a/MinimedKit/PumpModel.swift
+++ b/MinimedKit/PumpModel.swift
@@ -41,6 +41,12 @@ public enum PumpModel: String {
         return generation >= 23
     }
     
+    // On newer pumps, square wave boluses are added to history on start of delivery, and updated in place
+    // when delivery is finished
+    public var appendsSquareWaveToHistoryOnStartOfDelivery: Bool {
+        return generation >= 23
+    }
+    
     var hasLowSuspend: Bool {
         return generation >= 51
     }
--- a/RileyLinkKit/PumpOpsSynchronous.swift
+++ b/RileyLinkKit/PumpOpsSynchronous.swift
@@ -335,6 +335,12 @@ class PumpOpsSynchronous {
 
         var events = [TimestampedHistoryEvent]()
         var timeAdjustmentInterval: NSTimeInterval = 0
+        
+        // Going to scan backwards in time through events, so event time should be monotonically decreasing
+        // Start with some time in the future, to account for the condition when the pump's clock is ahead
+        // of ours by a small amount.
+        var timeCursor = NSDate(timeIntervalSinceNow: NSTimeInterval(minutes: 60))
+        let eventTimestampDeltaAllowance = NSTimeInterval(minutes: 60)
 
         pages: for pageNum in 0..<16 {
             NSLog("Fetching page %d", pageNum)
@@ -362,7 +368,11 @@ class PumpOpsSynchronous {
                         if date.compare(startDate) == .OrderedAscending  {
                             NSLog("Found event (%@) before startDate(%@)", date, startDate);
                             break pages
+                        } else if date.timeIntervalSinceDate(timeCursor) > eventTimestampDeltaAllowance {
+                            NSLog("Found event (%@) out of order in history. Ending history fetch.", date)
+                            break pages
                         } else {
+                            timeCursor = date
                             events.insert(TimestampedHistoryEvent(pumpEvent: event, date: date), atIndex: 0)
                         }
                     }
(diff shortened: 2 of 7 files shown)


RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -335,6 +335,12 @@ class PumpOpsSynchronous {
 
         var events = [TimestampedHistoryEvent]()
         var timeAdjustmentInterval: NSTimeInterval = 0
+        
+        // Going to scan backwards in time through events, so event time should be monotonically decreasing
+        // Start with some time in the future, to account for the condition when the pump's clock is ahead
+        // of ours by a small amount.
+        var timeCursor = NSDate(timeIntervalSinceNow: NSTimeInterval(minutes: 60))
+        let eventTimestampDeltaAllowance = NSTimeInterval(minutes: 60)
```
loudnate: Minor, but you might as well use this in line 342

ps2: I though so, but they're different things.  Line 342 is to handle pump vs phone differences (might need to increase this?), and line 343 is for variability between pump events (maybe should decrease this?); I think it's possible for events to appear out of order, for those events that get updated after the fact, like square wave boluses.

ps2: Hmm, now that I'm thinking about it, a square wave can go out for 8 hours... going to test and see how this looks in history.

loudnate: Square wave can last for 8 hours, but you're prevented from doing much else until it's done.

loudnate: I wonder if BasalProfileStart appears in those cases?

loudnate: It should be totally fine to say event start dates go backward within 1 hour if you test the scheduled event entry cases: BasalProfileStart/Sarah6e while a square is running. As long as the square isn't re-written as its delivered, after those events, then you're OK.

ps2: Here's what my testing (on a 522) shows:
1) Square wave boluses don't get added into the history until they are done being delivered
2) When they are added, they have their _start_ time as the timestamp
3) Other events can be added to history (changing alarm settings, etc) while the square wave is going
4) The ordering of events in the history is the order in which they are added, so temp basals end up with a timestamp that is out of order, and can be up to 8 hours out of order.

So I think I need to adjust the logic. This also has some implications for Looping and reservoir vs history processing.   From a history view, you don't know if a square wave is being delivered.

loudnate: Are you sure it doesn't show up during delivery? I believe it does on the 523. 

Re: reservoir, LoopKit always prefers reservoir over history unless the reservoir data is deemed incomplete. Even so, since the square wave prevents temp basal adjustments, it wouldn't be able to make changes based on incorrect data.

ps2: Yes, it looks like you're right about newer models.  Here's what I see on the 551:

1) Square wave boluses get added to history immediately upon start. The amount is 0. 
2) Events can be added while square wave is delivering.
3) Square wave record in history does not update until delivery is finished (or pump is suspended).  At that point, the amount is updated to reflect whatever amount has been delivered.
4) The updated record is still in the same place it was, so I don't see an out of order situation.

So it looks like out of order history events might only be a problem for 822s.  I'm wondering if there are any other records we may want to look into.  If not, I can adjust `eventTimestampDeltaAllowance` based on pump model.

loudnate: The only known mutable records, I believe, are Normal Bolus and Square
Bolus.

loudnate:
LGTM

RileyLinkKit/PumpOpsSynchronous.swift
ps2: Ok, Normal Bolus events on a 551 appear to work like Square Waves on older pumps; they're added to history at the _end_ of delivery, but have a start time of start of delivery, so out of order events are possible on newer pumps.   Looks like delivery is 1U/10s, and max bolus is 25U, so we have a max out of order of just over 4 minutes.

MinimedKit/PumpModel.swift
```diff
@@ -41,6 +41,12 @@ public enum PumpModel: String {
         return generation >= 23
     }
     
+    // On newer pumps, square wave boluses are added to history on start of delivery, and updated in place
+    // when delivery is finished
+    public var appendsSquareWaveToHistoryOnStartOfDelivery: Bool {
```
loudnate, line 46: Nice, we can use this in the `isMutable` calculation in the future.

PR #214, opened 2016-09-01
Improvements to wake method

ps2:
Avoid long (sometimes a minute or more) periods of spamming RF caused by many sessions queued up each individually trying to wake the pump.

This can happen when the pump is out of tune with RL, or if the pump is out of range.

With these changes, the queued up sessions skip over the 15s wake cycle, and quickly try to do their comms, error out, and move on.

### DIFF
--- a/MinimedKit/PumpModel.swift
+++ b/MinimedKit/PumpModel.swift
@@ -47,6 +47,10 @@ public enum PumpModel: String {
         return generation >= 23
     }
     
+    public var hasMySentry: Bool {
+        return generation >= 23
+    }
+    
     var hasLowSuspend: Bool {
         return generation >= 51
     }
--- a/RileyLinkBLEKit/RileyLinkBLEDevice.m
+++ b/RileyLinkBLEKit/RileyLinkBLEDevice.m
@@ -464,7 +464,7 @@
 }
 
 - (void) onIdle {
-    if (!runningIdle && idleListeningEnabled && _peripheral.state == CBPeripheralStateConnected && dataCharacteristic != nil) {
+    if (idleListeningEnabled && _peripheral.state == CBPeripheralStateConnected) {
         runningIdle = YES;
         NSLog(@"Starting idle RX");
         GetPacketCmd *cmd = [[GetPacketCmd alloc] init];
@@ -492,7 +492,7 @@
     if (idleListeningEnabled && !runningSession) {
         NSTimeInterval resetIdleAfterInterval = 2.0 * (float)_idleTimeout / 1000.0;
         
-        if (([NSDate dateWithTimeIntervalSinceNow:-resetIdleAfterInterval] > _lastIdle)) {
+        if (!runningIdle || ([NSDate dateWithTimeIntervalSinceNow:-resetIdleAfterInterval] > _lastIdle)) {
             [self onIdle];
         }
     }
--- a/RileyLinkBLEKit/RileyLinkBLEDevice.m
+++ b/RileyLinkBLEKit/RileyLinkBLEDevice.m
@@ -278,7 +278,7 @@
         NSString *foundVersion;
         BOOL versionOK = NO;
         // We run two commands here, to flush out responses to any old commands
-        [s doCmd:cmd withTimeoutMs:1000];
+        [s doCmd:cmd withTimeoutMs:5000];
         if ([s doCmd:cmd withTimeoutMs:1000]) {
             foundVersion = [[NSString alloc] initWithData:cmd.response encoding:NSUTF8StringEncoding];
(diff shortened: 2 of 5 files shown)


loudnate:
re-run CI?

MinimedKit/PumpModel.swift
```diff
@@ -47,6 +47,10 @@ public enum PumpModel: String {
         return generation >= 23
     }
     
+    public var hasMySentry: Bool {
```
loudnate, line 50: there might be a few places in code (RileyLinkBLEKit? Loop? I don't remember) that can be updated to take advantage of this instead of using `larger` or something else as a proxy...

RileyLinkBLEKit/RileyLinkBLEDevice.m
```diff
@@ -278,7 +278,7 @@ - (void)checkVersion {
         NSString *foundVersion;
         BOOL versionOK = NO;
         // We run two commands here, to flush out responses to any old commands
-        [s doCmd:cmd withTimeoutMs:1000];
+        [s doCmd:cmd withTimeoutMs:5000];
```
loudnate, line 281: Why the longer timeout here?

RileyLinkBLEKit/RileyLinkBLEDevice.m
```diff
@@ -464,7 +464,7 @@ - (void) setCustomName:(nonnull NSString*)customName {
 }
 
 - (void) onIdle {
-    if (idleListeningEnabled && _peripheral.state == CBPeripheralStateConnected) {
+    if (!runningIdle && idleListeningEnabled && _peripheral.state == CBPeripheralStateConnected && dataCharacteristic != nil) {
```
loudnate: The problem with requiring `!runningIdle` here is that it's sometimes never reset to `NO` (like if the RL goes out-of-range before the idle timeout). It's the explicit reason why line 495 is an OR:

``` objc
!runningIdle || ([NSDate dateWithTimeIntervalSinceNow:-resetIdleAfterInterval] > _lastIdle
```

The second condition is met when the system "misses" the timeout.

loudnate:
> caused by many sessions queued up 

what's the case where multiple sessions can queue up faster than they can run?

RileyLinkBLEKit/RileyLinkBLEDevice.m
ps2: I was seeing occasional timeouts on that command.  I've been trying to eliminate all the RileyLinkTimeouts, as they shouldn't really occur unless something is wrong with the connection to the RileyLink.   As this only happens on startup, and normally returns fast, it shouldn't have any negative side effects, and it may give the OS more time to set up all the bluetooth stuff.  If we do get a timeout here, it not flushing as intended, and the get version was failing.

RileyLinkBLEKit/RileyLinkBLEDevice.m
ps2: Yeah, I'm trying to fix too many things in this PR; I just keep seeing issue after issue when watching the logs closely. I'll back this part out and see if I can re-create the issue this was intending to solve.

RileyLinkBLEKit/RileyLinkBLEDevice.m
loudnate: Gotcha. I thought I remembered this failing only when there's a mismatch in the command-response conversation because of a sudden re-launch of the app. I didn't think a timeout could help that issue, but you're right that it can't hurt either.

loudnate:
LGTM

RileyLinkBLEKit/RileyLinkBLEDevice.m
loudnate: What was the issue, do you remember?

ps2: It was another getting out of sync issue, I believe; it may have even been a situation where it was repeatedly thinking it was getting interrupted, but the idle loop was interrupting itself.  I was doing a bunch of testing with the pump out of range to see what kinds of situations arose.  If it's an issue, I'm sure I'll be able to repeat it, and look at it more closely.  I was going too fast, trying to wrap things up last night.

ps2:
RE: sessions queuing up: In the RL app, it fetches status and history back to back (if they're both needed). I was also issuing commands from the UI on top of that. 

But even if you had just two sessions queued, and they both did 3 retries of the wakeup cycle (16s) that's a minute and half of spamming RF.

loudnate:
Gotcha. 

As we understand the sync issues more (I only have real-world observations) I could express them as unit tests.

PR #222, opened 2016-09-05
Allow rate and duration records for a temp basal to be out of order, …

ps2:
…and verify timestamp of events


loudnate:
LGTM, thanks!

ps2:
I think you're right that a better way is to allow clients to upload NS treatments, so we'll have to tackle that soon.

PR #224, opened 2016-09-05
v0.11.2

ps2:
- Nightscout upload handles out of order temp basal rate & duration events now, and verifies timestamps when matching them up.


loudnate:
LGTM

PR #225, opened 2016-09-09
Adding ChangeMeterID and DailyTotal515 pump events

ps2:
cc @bewest and @ecc1

### DIFF
--- a/MinimedKit/PumpEvents/DailyTotal523PumpEvent.swift
+++ b/MinimedKit/PumpEvents/DailyTotal523PumpEvent.swift
@@ -1,5 +1,5 @@
 //
-//  Sara6EPumpEvent.swift
+//  DailyTotal523PumpEvent.swift
 //  RileyLink
 //
 //  Created by Pete Schwamb on 3/8/16.
@@ -8,7 +8,7 @@
 
 import Foundation
 
-public struct Sara6EPumpEvent: PumpEvent {
+public struct DailyTotal523PumpEvent: PumpEvent {
     
     public let length: Int
     public let rawData: NSData
@@ -33,7 +33,7 @@ public struct Sara6EPumpEvent: PumpEvent {
     
     public var dictionaryRepresentation: [String: AnyObject] {
         return [
-            "_type": "Sara6E",
+            "_type": "DailyTotal523",
             "validDate": validDateStr,
         ]
     }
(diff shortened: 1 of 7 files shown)


MinimedKit/PumpEvents/DailyTotal523PumpEvent.swift
```diff
@@ -33,7 +33,7 @@ public struct Sara6EPumpEvent: PumpEvent {
     
     public var dictionaryRepresentation: [String: AnyObject] {
         return [
-            "_type": "Sara6E",
+            "_type": "DailyTotal523",
```
loudnate, line 36: Nice, though I'll always remember this as Sara6e. I want to get around decoding this event, since this sort of daily-summary data is what I'd like to report on long-term. I was thinking of setting Loop to download history once daily even if the preference is set to Reservoir, just for archive/analysis.

Do you have a recent carelink pull you wouldn't mind contributing to that effort, possibly for the unit test case below?

loudnate:
LGTM. I had a laugh today looking at the Travis CI macOS load. Release week is always ridiculous.

MinimedKit/PumpEvents/DailyTotal523PumpEvent.swift
ps2: There are three different daily totals types for different pumps. Unfortunately, I don't have any Carelink records for that particular date.

ecc1, line 36: On Thu, Sep 08, 2016 at 06:22:01PM -0700, Nathan Racklyeft wrote:

> Do you have a recent carelink pull you wouldn't mind contributing to that
> effort, possibly for the unit test case below?

That's odd: I have Carelink CSV files and corresponding history page
dumps for some periods where those records occur, but the CSV files
don't list them at all.  It just goes from the previous
BasalProfileStart, to the ResultDailyTotal, to the 00:00
BasalProfileStart.  The history page dumps have them though, right
before the ResultDailyTotal record.

## 

Eric Cooper             e c c @ c m u . e d u

ps2: Yeah, I did some searching through the CSVs and I don't see anything that looks like the daily summary either.

loudnate, line 36: You're right, I thought I remember a midnight event but all I see summary-wise is what shows up in the CSV at the time of the history dump (today units delivered, yesterday units delivered).

PR #227, opened 2016-09-09
Tune updates

ps2:
The pump keeps listening for over a minute after the last comms.  We can avoid sending out hundreds of wake packets by doing a small, 1-2 packet model check to see if the pump is still listening, and extend the wake timeout window. 

Also included in the PR are a couple changes for tuning:  if we need to tune, start in the middle of the frequency range, to increase our chances of waking the pump, and if we fail to find the pump, don't leave the frequency at the end of the range.

### DIFF
--- a/RileyLinkKit/PumpOpsSynchronous.swift
+++ b/RileyLinkKit/PumpOpsSynchronous.swift
@@ -421,8 +421,8 @@ class PumpOpsSynchronous {
             results.bestFrequency = sortedTrials.first!.frequencyMHz
             try setBaseFrequency(results.bestFrequency)
         } else {
-            throw PumpCommsError.RFCommsFailure("No pump responses during scan")
             try setBaseFrequency(middleFreq)
+            throw PumpCommsError.RFCommsFailure("No pump responses during scan")
         }
         
         return results
--- a/RileyLinkKit/PumpOpsSynchronous.swift
+++ b/RileyLinkKit/PumpOpsSynchronous.swift
@@ -138,7 +138,7 @@ class PumpOpsSynchronous {
             let msg = makePumpMessage(.GetPumpModel)
             let response = try sendAndListen(msg, retryCount: 1)
             
-            if response.messageType == .GetPumpModel && (response.messageBody as? GetPumpModelCarelinkMessageBody) != nil {
+            if response.messageType == .GetPumpModel && response.messageBody is GetPumpModelCarelinkMessageBody {
                 return true
             }
         } catch {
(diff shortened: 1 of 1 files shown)


RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -132,12 +132,35 @@ class PumpOpsSynchronous {
             throw lastError
         }
     }
+    
+    private func pumpResponding() -> Bool {
+        do {
+            let msg = makePumpMessage(.GetPumpModel)
+            let response = try sendAndListen(msg, retryCount: 1)
+            
+            if response.messageType == .GetPumpModel && (response.messageBody as? GetPumpModelCarelinkMessageBody) != nil {
```
loudnate: `(response.messageBody as? GetPumpModelCarelinkMessageBody) != nil` is equivalent to `response.messageBody is GetPumpModelCarelinkMessageBody`

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -132,12 +132,35 @@ class PumpOpsSynchronous {
...
+
 
     private func wakeup(duration: NSTimeInterval = NSTimeInterval(minutes: 1)) throws {
         guard !pump.isAwake else {
             return
         }
         
+        if pumpResponding() {
+            NSLog("Pump responding despite our wake timer having expired. Extending timer")
+            // By my observations, the pump stays awake > 1 minute past last comms. Usually
```
loudnate, line 158: really? You're suggesting that comms have a side effect of extending the wakeup time?

ps2: Yes, it certainly does.  I saw evidence of this even before hooking up the power monitor.  Just never measured it and took advantage of it in code.

loudnate: wow, just when I think I understand this device...

loudnate:
LGTM

RileyLinkKit/PumpOpsSynchronous.swift
loudnate: Any risk of this being true only for the x23+ pumps?

ps2: Most of my testing the other day was on the 515WW.  I should maybe test more on 522 and 523, but I know I've seen it on 523s hooked up to my bench supply.

ps2: Just tested again on a 523.  Waited for the app to wake it up, then waited 40s and did a fetch model, then turned off the RL to ensure no more comms. The pump continued to use power (which is indicative of RX) for 1m45s. 

The "Pump responding despite our wake timer having expired. Extending timer" message also is seen in the logs fairly frequently on the 523 while I've been playing around.

loudnate: 👍👍👍

PR #231, opened 2016-09-12
Swift 3: NightscoutUploadKit, Crypto, RileyLink

loudnate:


loudnate:
Tested branch on a device running iOS 9.3.5, was able to run pump commands. I couldn't test Nightscout auth or uploading though.

PR #232, opened 2016-09-12
Delete Cartfile.resolved

beached:
The brand on loudnate/Crypto this was built against is no longer there.  It will fail for someone without the carthage cache file.  I deleted this file and ran boostrap and it built


loudnate:
Can you check-in the newly-created version of your `Cartfile.resolved`?

PR #234, opened 2016-09-12
Use Crypo framework from soffes repo

ps2:


loudnate:
Yep, that's exactly what I was going to do :)

Cartfile
```diff
@@ -1 +1 @@
-github "loudnate/Crypto" "master"
+github "soffes/Crypto" "master"
```
loudnate: s/`"master"`/`~> 0.4.0` would probably be better.

ps2:
And travis CI is happy!

ps2:
Thanks @beached and @mhaeberli for looking into this.

loudnate:
LGTM

PR #240, opened 2016-09-18
Added Pump Self Test History Event

beached:


MinimedKit/PumpEvents/SelfTestPumpEvent.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  SelfPumpEvent.swift
```
loudnate: No need to create this file, actually, since it's a basic 7-byte record

MinimedKit/PumpEventType.swift
```diff
@@ -112,6 +113,8 @@ public enum PumpEventType: UInt8 {
             return JournalEntryPumpLowBatteryPumpEvent.self
         case .battery:
             return BatteryPumpEvent.self
+        case .selftest:
```
loudnate: This case isn't needed either

loudnate:
Thank you!!

loudnate:
Can try building this? Looks like CI couldn't.

loudnate:
Ah you're missing project file changes. If you remove the new file, that won't be necessary anyway though.

beached:
Oh yah, I forgot about the generic one.  Ill fix

PR #241, opened 2016-09-18
Added Pump Self Test History Event

beached:
This time uses the placeholder event type as it is only a timestamped event


loudnate:
Thanks!

loudnate (approved):

PR #242, opened 2016-09-18
Version 0.12.1

loudnate:
- Fixes MySentry pairing
- Adds support for the SelfTest pump event (0x20)


ps2:
Looks great to me.  Thanks for dealing with these updates while I was away!

Issue #243, opened 2016-09-22
Out-of-order history events should be returned


loudnate:
![img_1254](https://cloud.githubusercontent.com/assets/212876/18769022/70a534fa-80de-11e6-908d-86c3c2d78685.PNG)

Here's an example of what we recently discovered: bolus events are written after completion, and potentially out-of-order, for x22 pumps.

We still need to change [this condition](https://github.com/ps2/rileylink_ios/blob/dev/RileyLinkKit/PumpOpsSynchronous.swift#L483), I believe.

ps2:
I think you're right about that condition changing.  If we change it, though, we'll be returning the same items repeatedly from subsequent calls of this method; we may need to add some de-duping code.

loudnate:
We already have to dedupe a small number of records on every call, because we use >= the last record date. 

LoopKit is using a unique constraint on (date, raw data), which handles the current cases (I believe).

loudnate:
If we need to minimize the repetition, we could consider a property on the PumpEvent protocol called "canBeOutOfOrder" and only return those

ps2:
I start to wonder if tracking by page # and record index is a better idea than time.  If we could track that, we'd only have to track as far back as the last modifiable record (which only exist on x23 and newer pumps).

loudnate:
Isn't the most recent page always page 0? CGM pages truly ring, but I don't think history does (externally).

ps2:
yes, but I think there is a getPageNumber command that returns a counter that you can use to detect rollovers.  We could also generate a page hash with the first record in the page.

ps2:
I think this might suffice: 6a024e6

loudnate:
> yes, but I think there is a getPageNumber command that returns a counter that you can use to detect rollovers. We could also generate a page hash with the first record in the page.

excellent – you're sure that `getPageNumber` isn't returning the glucose page number, right?

PR #244, opened 2016-09-23
return events that may be timestamped older, but were actually recent…

ps2:
…ly added.


ps2:
Fixes #243

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -481,7 +481,9 @@ class PumpOpsSynchronous {
                             NSLog("Found event (%@) out of order in history. Ending history fetch.", date as NSDate)
                             break pages
                         } else {
-                            timeCursor = date
+                            if (date.compare(startDate) == .orderedAscending) {
```
loudnate: This needs to be `!=`?

loudnate (approved):
LGTM. Might be worth an x22 user verifying the deduping behavior before a general release.

PR #246, opened 2016-09-23
Add resetEvent

ps2:


ps2:
This is invalid. There is a different reason this "record" was showing up.

PR #247, opened 2016-09-24
Breaks up long history entries into 1023 chars

beached:
There is a limit in ios10 of 1k for NSLog entries.  This will ensure that the history pages are fullly displayed

### DIFF
--- a/RileyLinkKit/PumpOpsSynchronous.swift
+++ b/RileyLinkKit/PumpOpsSynchronous.swift
@@ -464,7 +464,17 @@ class PumpOpsSynchronous {
                     throw error
                 }
             }
-            
+            if true {
+                let logString = NSString(format: "Fetched page %d: %@", pageNum, pageData as NSData ) 
+                let sz = logString.length
+                var idx = 0 
+                while idx < sz {
+                    let jmp = sz - idx > 1023 ? 1023 : sz - idx 
+                    NSLog( logString.substring(with: NSRange( location:idx, length: jmp ))) 
+                    idx += jmp 
+                }   
+            }   
+
             NSLog("Fetched page %d: %@", pageNum, pageData as NSData)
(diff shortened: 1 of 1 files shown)


RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -464,8 +464,17 @@ class PumpOpsSynchronous {
                     throw error
                 }
             }
-            
-            NSLog("Fetched page %d: %@", pageNum, pageData as NSData)
+            if true {
```
loudnate, line 467: Remove this?

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -464,8 +464,17 @@ class PumpOpsSynchronous {
...
+                let logString = NSString(format: "Fetched page %d: %@", pageNum, pageData as NSData ) 
+                let sz = logString.length
```
loudnate, line 469: Iterating over a string is thorny due to multibyte character support. Iterating over `Data` is simple. Could you change this loop to iterate over blocks of `pageData` instead?

Also, the length of `String(describing: pageData)` is always 2,304 characters, which you could rely on.

loudnate (changes requested):

RileyLinkKit/PumpOpsSynchronous.swift
ps2: It should always be 1024 bytes.   4 lines of 256 bytes would be great (512 characters/line + some prefix to separate it from the rest of the log)
Something like `page 1 - bytes (000-255)` would be a nice prefix

loudnate: I like that.

(my character number was based on a space after every 4 bytes `1024 * 2 + (1024 / 4)`)

ps2:
Added above suggestions in #253

PR #248, opened 2016-09-24
deduplicate history events before returning

loudnate:


RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -468,8 +468,13 @@ class PumpOpsSynchronous {
             NSLog("Fetched page %d: %@", pageNum, pageData as NSData)
             let page = try HistoryPage(pageData: pageData, pumpModel: pumpModel)
 
+            // Prevent returning duplicate content, which is possible e.g. in the case of rapid RF temp basal setting
+            var seenEventData = Set<Data>()
+
```
ps2, line 473: Would you want to move this out of the per-page scope to cover dups across pages? I know it's probably a slim chance, but...

loudnate:
Oh yeah, absolutely! 

I'm out of town until Monday evening, so feel free to make that change if you want to release it sooner.

PR #249, opened 2016-09-24
Read settings

ps2:
Expose a command read pump settings, like max bolus, max temp basal.

### DIFF
--- a/RileyLinkKit/PumpOps.swift
+++ b/RileyLinkKit/PumpOps.swift
@@ -55,13 +55,13 @@ public class PumpOps {
         }
     }
     
-    public func readSettings(_ completion: @escaping (Either<String, Error>) -> Void)  {
+    public func readSettings(_ completion: @escaping (Either<ReadSettingsCarelinkMessageBody, Error>) -> Void)  {
         device.runSession(withName: "Read pump settings") { (session) -> Void in
             let ops = PumpOpsSynchronous(pumpState: self.pumpState, session: session)
             do {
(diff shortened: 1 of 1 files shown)


RileyLinkKit/PumpOps.swift
```diff
@@ -55,6 +55,23 @@ public class PumpOps {
         }
     }
     
+    public func readSettings(_ completion: @escaping (Either<ReadSettingsCarelinkMessageBody, Error>) -> Void)  {
+        device.runSession(withName: "Read pump settings") { (session) -> Void in
+            let ops = PumpOpsSynchronous(pumpState: self.pumpState, session: session)
+            do {
+                let response: ReadSettingsCarelinkMessageBody = try ops.messageBody(to: .readSettings)
+                DispatchQueue.main.async { () -> Void in
```
loudnate, line 63: Necessary, or just habit?

ps2: Are you referring to the dispatch on main thread? I thought I'd fit the pattern and keep consistent behavior. 

Or are you asking if I actually need to expose readSettings?  I thought it might be useful if we want to diagnose temp basal failures, and for setting Loop configuration from pump.

loudnate: I meant the main queue dispatch, since the vast majority of callers are invoking PumpOps from a background queue anyway.

There's not currently a consistent behavior in these commands, by my count: 4.5 commands do dispatch to main and the other 8 don't.

loudnate (approved):

PR #253, opened 2016-09-24
Fix history page logging truncation

ps2:
Thanks @beached!


loudnate:
Thanks! I know it can be hard to use the new Swift types.

PR #254, opened 2016-09-25
Use high bits of temp basal to correctly decode higher temp basal rates

ps2:
Decodes rates up to 34U/hr, which is the pump's max. 
Previously, anything higher than 6.375 was being decoded incorrectly.


ps2:
cc @bewest @ecc1

loudnate (approved):

PR #265, opened 2016-10-02
Making sure to only used fixed-offset timezones

loudnate:
I should have done this from the beginning. Once a timezone is serialized to defaults, it became fixed, but this will handle the case of multi-plane, multi-timezone, multi-time change trips.


loudnate:
Alright – I've tested this both with existing installs and new installs. Going to merge down to make the 1.0 release of Loop.

PR #267, opened 2016-10-09
Add missing 7xx models

ps2:


christerjensen (approved):
My 715 WW pump now sends recent history and pump status to Rileylink iOs. Previously it only sent pump model and accepted Push button

Issue #200, opened 2016-08-17
Ability to use MM CGM with an x22 Pump


tomasboudr:
Looking for the ability to use a MM CGM with an x22 pump, thus we will need to parse sensor data pages in rileylink

Related issue in Loop: https://github.com/loudnate/Loop/issues/100

ps2:
See getHistoryEvents in https://github.com/ps2/rileylink_ios/blob/dev/RileyLinkKit/PumpOpsSynchronous.swift for an example of the swift code you'll need to write to fetch the pages

ps2:
@bewest  has Python code to decode CGM pages here: https://github.com/openaps/decocare/blob/master/decocare/cgm/__init__.py

tmecklem:
I'm spiking a solution to this at https://github.com/tmecklem/rileylink_ios/tree/support-paged-cgm.

ps2:
Support for fetching CGM pages has been merged in.  Still not uploading CGM data to NS.

ps2:
Fixed with #275

PR #270, opened 2016-10-20
Support paged cgm

tmecklem:
Summary: Add support for fetching and displaying recent data from cgm pages. See #200 

This PR adds to MinimedKit the ability to fetch and decode cgm pages into events similar to the way that history pages are processed. It also supports fetching the pointer to the current page of cgm data to support fetching the most recent pages first.

In RileyLinkKit, functionality is built upon the additions to MinimedKit to fetch recent cgm data from the last 24 hours and display it similarly to the fetch recent history functionality.


MinimedKit/Extensions/NSData.swift
```diff
@@ -71,6 +71,14 @@ extension Data {
      */
 }
 
+extension Data {
+    func reverseBytes() -> Data {
```
ps2: There is a method in the standard library to do this already.  See https://developer.apple.com/reference/foundation/data/1780245-reversed

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -544,6 +544,106 @@ class PumpOpsSynchronous {
...
+    
+    internal func getGlucoseHistoryEvents(since startDate: Date) throws -> ([TimestampedGlucoseEvent], PumpModel) {
+        try wakeup()
+        
+        let pumpModel = try getPumpModel()
+        
+        var events = [TimestampedGlucoseEvent]()
+        
+        // Going to scan backwards in time through events, so event time should be monotonically decreasing.
+        let eventTimestampDeltaAllowance = TimeInterval(hours: 9)
```
ps2: This should not be necessary, unless you have encountered out of order glucose events, and need to keep looking back further in history to ensure you get all events since startDate.

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -544,6 +544,106 @@ class PumpOpsSynchronous {
...
+            var idx = 0
+            let chunkSize = 256;
+            while idx < pageData.count {
+                let top = min(idx + chunkSize, pageData.count)
+                let range = Range(uncheckedBounds: (lower: idx, upper: top))
+                NSLog(String(format: "GlucosePage %02d - (bytes %03d-%03d): ", pageNum, idx, top-1) + pageData.subdata(in: range).hexadecimalString)
+                idx = top
+            }
+            
+            let page = try GlucosePage(pageData: pageData, pumpModel: pumpModel)
```
ps2: I searched and didn't see any actual dependency on pumpModel in the data parsing.  Can this be removed?

ps2 (changes requested):
Amazing work, Tim! Looking forward to getting this into Loop.  It would also be nice to add functionality for CGM upload to NS for those users using just the RL app, though I think that's a smaller set.

RileyLinkKit/PumpOpsSynchronous.swift
tmecklem: Good catch! 👍

RileyLinkKit/PumpOpsSynchronous.swift
tmecklem: There are out of order events due to relative timestamping. I changed the timestamp check on https://github.com/ps2/rileylink_ios/pull/270/commits/f266ebbf41e9888470d9db92285018331b2afda4#diff-bf2d04fbde756ea81dd46dafafbdacddR589 to use the reference timestamped events only. That should ensure that all relevant glucose events are captured since startDate at the expense of maybe capturing a few extra independently timestamped (non-sgv) entries.

MinimedKit/Extensions/NSData.swift
tmecklem: I'm having trouble figuring out what to do with the data structure that reversed() gives me. It's a `ReversedRandomAccessCollection<Data>`, and it looks like I have to use a different Index structure. The change seems to leak all the way through to how I access rawData GlucoseEvent and seems messy. Can you take a look and help me understand how to appropriately use reversed() on the page Data?

ps2: You can use Data(data.reversed()) to create a new reversed data object.  There may be ways to work with RandomAccessCollection<Data> types, and avoid the copy, but I wasn't able to get that working in my swift playground.

RileyLinkKit/PumpOps.swift
```diff
@@ -144,6 +144,34 @@ public class PumpOps {
...
+     
+     History timestamps are reconciled with UTC based on the `timeZone` property of PumpState, as well as recorded clock change events.
+     
+     - parameter startDate:  The earliest date of events to retrieve
+     - parameter completion: A closure called after the command is complete. This closure takes a single Result argument:
+     - success(events): An array of fetched history entries, in ascending order of insertion
+     - failure(error):  An error describing why the command failed
+     
+     */
+    public func getGlucoseHistoryEvents(since startDate: Date, completion: @escaping (Either<(events: [TimestampedGlucoseEvent], pumpModel: PumpModel), Error>) -> Void) {
```
ps2: Does pumpModel need to be here?

MinimedKit/Extensions/NSData.swift
tmecklem: Thanks! Change made.

tmecklem:
I'm planning to get this version of the app installed on the phone of someone using a 722 pump with enlite before the start of the weekend for Real World:tm: testing.

RileyLinkKit/PumpOps.swift
tmecklem: Nope!

ps2 (approved):

PR #273, opened 2016-10-23
Fix timestamp and paging

tmecklem:
- Fixes a bug that causes glucose readings to be off due to incorrect relative timestamps applied to "19-Something" opcode records in the page.
- Looks back more than 1 page of history and correctly handles reading page 0


ps2 (approved):
LGTM!

PR #276, opened 2016-11-23
add BatteryChemistryType.swift

ps2:


MinimedKit/BatteryChemistryType.swift
```diff
@@ -0,0 +1,46 @@
...
+    var minVoltage: Double {
+        switch self {
+        case .alkaline:
+            return 1.26
+        case .lithium:
+            return 1.32
+        }
+    }
+
+    func chargeRemaining(voltage: Double) -> Double {
```
loudnate, line 42: Nit: `voltage voltage` would keep the parameter name from being implicit to callers.

MinimedKit/BatteryChemistryType.swift
```diff
@@ -0,0 +1,46 @@
...
+//  BatteryChemistryType.swift
+//  RileyLink
+//
+//  Created by Jeremy Lucas on 11/15/16
+//  Copyright © 2016 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
+enum BatteryChemistryType: Int, CustomStringConvertible {
```
loudnate, line 11: `public` (same for any methods/properties you want below)

MinimedKit/BatteryChemistryType.swift
ps2: Not sure I follow; right now it is explicit: i.e. callers must specify `voltage:` as the label when calling. Are you saying to do `_ voltage:` to make it implicit?

PR #288, opened 2016-12-04
Fixed ChangeTime History Event

beached:
The ChangeTime history event was thought to be 14bytes long but is in fact 7.  It was almost always followed by another 7 byte event, NewTime.  In some cases, on History Page boundaries these packets are not beside each other.  

This PR will remove the last 7 bytes of the ChangeTime definition and remove the test that assumed that.

The code that calculates time adjustment now assumes two consecutive events(with or without history page boundaries) and subtracts the timestamp of ChangeTime from that of NewTime to get the adjustment number.


ps2:
Looks like there are some other failing tests, because the offsets changed:
```
/Users/travis/build/ps2/rileylink_ios/MinimedKitTests/HistoryPageTests.swift:347: error: -[MinimedKitTests.HistoryPageTests testBolusReminderAndReturnError:] : XCTAssertEqual failed: ("29") is not equal to ("32") - 
Could not cast value of type 'MinimedKit.BasalProfileStartPumpEvent' (0x10da54c30) to 'MinimedKit.BolusReminderPumpEvent' (0x10da58538).
```

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -501,9 +502,12 @@ class PumpOpsSynchronous {
                     }
                 }
 
-                if let event = event as? ChangeTimePumpEvent {
-                    timeAdjustmentInterval += event.adjustmentInterval
+                if let event = event as? NewTimePumpEvent {
+                    if let ct_event = lastEvent as? ChangeTimePumpEvent {
```
ps2: You don't need two separate if statements here, you can have two lets separated by a comma.

ps2 (changes requested):

RileyLinkKit/PumpOpsSynchronous.swift
ps2: Also, ct_event should be ctEvent

ps2:
https://github.com/LoopKit/Loop/issues/304

ps2 (approved):

PR #293, opened 2016-12-10
Fix x22 cgm page gaps

tmecklem:
Changes to enhance the reliability and correctness of cgm page decoding

Details:

* Change event decoding to add sensor packets, sensor errors, low and high glucose data
* Remove 0x13 and 0x01 events incorrectly marked as offset-incrementing events
* Change so that only sensor timestamp events can serve as reference for relative offset events
* Time offset happens now in reverse chronological order from sensor timestamps (the same direction as decoding happens from end of page to beginning).
* cgm events at the end of the page that are not followed by a sensor timestamp receive their offset from the preceding sensor timestamp, unless the cgm reading event chain is broken by another type of event
* Add glucose history timestamp write command and page re-fetch, that is used when an appropriate reference timestamp is unavailable in the current cgm page.
* Additional attributes of known events are decoded, such as "page_end", "gap" or "last_rf" timestamp type on sensor timestamp.


MinimedKit/GlucoseEvents/SensorCalGlucoseEvent.swift
```diff
@@ -11,10 +11,10 @@ import Foundation
 public struct SensorCalGlucoseEvent: RelativeTimestampedGlucoseEvent {
     public let length: Int
     public let rawData: Data
-    public let waiting: String
-    public var timestamp: DateComponents
+    public let calibrationType: String
```
loudnate: If this value is useful to public clients, its options should be defined as an enum. If it's just for debugging in `dictionaryRepresentation`, then it should be made `private`.

This comment also applies similar approaches below: `errorType`, `packetType`, `statusType`

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -474,6 +478,18 @@ public class RileyLinkDeviceTableViewController: UITableViewController, TextFiel
                     }
                     return NSLocalizedString("Fetching glucose…", comment: "Progress message for fetching pump glucose.")
                 }
+            case .writeGlucoseHistoryTimestamp:
+                vc = CommandResponseViewController { [unowned self] (completionHandler) -> String in
+                    self.device.ops?.writeGlucoseHistoryTimestamp() { (response) -> Void in
+                        switch response {
+                        case .success(_):
+                            completionHandler(String(describing: "Okay, thing worked"))
```
loudnate: `String(describing: "Okay, thing worked")` is more concisely written as `"Okay, thing worked"`. Future users might thank you if this was more descriptive...

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
```diff
@@ -474,6 +478,18 @@ public class RileyLinkDeviceTableViewController: UITableViewController, TextFiel
...
+                        case .failure(let error):
+                            completionHandler(String(describing: error))
+                        }
+                    }
+                    return NSLocalizedString("Fetching glucose…", comment: "Progress message for fetching pump glucose.")
```
loudnate: This is copied from line 479

RileyLinkKit/UI/RileyLinkDeviceTableViewController.swift
tmecklem: LOL, didn't mean to leave that in there.

MinimedKit/GlucoseEvents/SensorCalGlucoseEvent.swift
tmecklem: So far, the additional information is useful for diagnosing problems but there's no different actions taken based on any of it. I'll change those to private.

PR #296, opened 2016-12-12
X22 sensor timestamp refinement

tmecklem:
Refines the cgm page timestamping and adds low and high sgv to nightscout

* Remove copy/paste write sensor timestamp repeatCount of 255
* Group sensor data, sensor high, and sensor low events under SensorDataGlucoseEvent for easier sgv filtering/processing to Nightscout and loop.
* Do not use "gap" sensor timestamps for forward offsets since they represent the start of a gap in time after the timestamp.
* Do not attempt to write a sensor timestamp when decoding a completed history page.


MinimedKit/GlucoseEvents/SensorTimestampGlucoseEvent.swift
```diff
@@ -8,11 +8,32 @@
...
 import Foundation
 
-public struct SensorTimestampGlucoseEvent: ReferenceTimestampedGlucoseEvent {
+public enum SensorTimestampType: String {
+    case lastRf
+    case pageEnd
+    case gap
+    case unknown
+    
+    public static func eventType(code: UInt8) -> SensorTimestampType {
```
loudnate: This would be more semantic as an initializer:
`public init(code: UInt8)`
(also it might not need to be public)

MinimedKit/GlucoseEvents/SensorTimestampGlucoseEvent.swift
```diff
@@ -27,20 +48,14 @@ public struct SensorTimestampGlucoseEvent: ReferenceTimestampedGlucoseEvent {
...
-        case 0x00:
-            timestampType = "last_rf"
-        case 0x01:
-            timestampType = "page_end"
-        case 0x02:
-            timestampType = "gap"
-        default:
-            timestampType = "unknown"
-        }
+        timestampType = SensorTimestampType.eventType(code: UInt8(d(3) >> 5) & 0b00000011)
```
loudnate: Sorry for just noticing this now, but "`d`" is useless here as you're undoing the unnecessary `Int` cast. Coupled with the initializer change: 

```swift
timestampType = SensorTimestampType(code: availableData[3] >> 5 & 0b11)
```

is much cleaner

MinimedKit/GlucosePage.swift
```diff
@@ -69,8 +68,10 @@ public class GlucosePage {
                 let event = matchEvent(tempOffset, relativeTimestamp: DateComponents())
                 if event is RelativeTimestampedGlucoseEvent {
                     relativeEventCount += 1
-                } else if let event = event as? ReferenceTimestampedGlucoseEvent {
-                    let offsetDate = calendar.date(byAdding: Calendar.Component.minute, value: 5 * relativeEventCount, to: event.timestamp.date!)!
+                } else if let sensorTimestampEvent = event as? SensorTimestampGlucoseEvent,
+                    relativeEventCount == 0 || (event as! SensorTimestampGlucoseEvent).isForwardOffsetReference() {
```
loudnate: Replace that last part with simply `sensorTimestampEvent.isForwardOffsetReference()` as your unwrapping statement is in-scope at this point.

MinimedKit/GlucosePage.swift
```diff
@@ -69,8 +68,10 @@ public class GlucosePage {
...
+                    
+                    let offsetDate = calendar.date(byAdding: Calendar.Component.minute, value: 5 * relativeEventCount, to: sensorTimestampEvent.timestamp.date!)!
```
loudnate: `byAdding: .minute,` since the type is inferred.

Also, I'm not sure this line to be a `calendar` operation if you're just adding a double to a double (the next one absolutely does)
`sensorTimestampEvent.timestamp.date?.addingTimeInterval(TimeInterval(minutes: 5 * relativeEventCount))`

MinimedKit/GlucosePage.swift
```diff
@@ -69,8 +68,10 @@ public class GlucosePage {
...
                     var relativeTimestamp = calendar.dateComponents([.year, .month, .day, .hour, .minute], from: offsetDate)
                     relativeTimestamp.calendar = calendar
```
loudnate: Are you sure you need this line?

MinimedKit/GlucoseEvents/SensorTimestampGlucoseEvent.swift
tmecklem: I wondered if there wasn't a better way. Thanks for the feedback on this.

MinimedKit/GlucosePage.swift
tmecklem: Which line are you referring to? 72?

MinimedKit/GlucosePage.swift
tmecklem: I thought I might be able to do that but Xcode was throwing a fit about it. Must have been a stale compiler error because it worked this time around. 👍

MinimedKit/GlucosePage.swift
loudnate: 75/76. I assumed that creating `DateComponents` from a `Calendar` would result in components that included a reference to the calendar.

tmecklem: Ah, good question. I'll try some things out.

MinimedKit/GlucosePage.swift
tmecklem: I'm not sure how to go about changing this. Could you help me out with an example of what you think could improve it?

tmecklem:
@loudnate, I appreciate you taking some of your time to review this code and make it better. Both because it's not a language I'm all that familiar with yet, and because you know this codebase so well. 👍

MinimedKit/GlucosePage.swift
loudnate: I thought I did?
```swift
let offsetDate = sensorTimestampEvent.timestamp.date!.addingTimeInterval(TimeInterval(minutes: 5 * relativeEventCount))
```
force-unwrapping should be avoided when possible, so I assume you're certain that `timestamp.date` is non-nil?

PR #316, opened 2017-03-20
Fetch fewer pages and reduce pump comms time when fetching event history

sharpfive:
**updated by ps2**
The current history fetching method searches 9 additional hours before startTime on x22 pumps, and 10 additional minutes on x23 pumps.  This is to deal with square waves that can be up to 8 hours out of order in pump history.  A better way to deal with that is to just ignore delayed append events when scanning to see if we've retrieved enough events.


RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -468,18 +468,24 @@ class PumpOpsSynchronous {
...
-                    } else if date.timeIntervalSince(timeCursor) > eventTimestampDeltaAllowance {
-                        NSLog("Found event (%@) out of order in history. Ending history fetch.", date as NSDate)
-                        return (events: events, hasMoreEvents: false)
-                    } else {
-                        if (date.compare(startDate) != .orderedAscending) {
-                            timeCursor = date
+                    
+                    let dontCancelBecauseOfDate = pumpModel.mayHaveOutOfOrderEvents && event.canBeDelayedAppend()
+                    
+                    if !dontCancelBecauseOfDate {
```
ps2: !dontCancel is confusing.

MinimedKit/PumpEvents/PumpEvent.swift
```diff
@@ -21,3 +21,16 @@ public protocol PumpEvent : DictionaryRepresentable {
     }
     
 }
+
+public extension PumpEvent {
+    public func canBeDelayedAppend() -> Bool {
```
ps2: Change this to mayBeOutOfOrderOnPumpModel(pumpModel)

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -179,8 +195,103 @@ class PumpOpsSynchronousTests: XCTestCase {
         // It should be included
         XCTAssertTrue(array(timeStampedEvents, containsPumpEvent: squareWaveBolus))
     }
+    
+    func testOutOfOrderEventFor522() {
+        loadTestWithPumpModel(.Model522)
+        
+        let tempEventBasal = createTempEventBasal()
```
ps2: To read this test easily, we need to know what time tempEventBasal is.

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -468,18 +468,24 @@ class PumpOpsSynchronous {
...
+                        if date.timeIntervalSince(startDate) < -eventTimestampDeltaAllowance {
+                            NSLog("Found event at (%@) to be more than %@s before startDate(%@)", date as NSDate, String(describing: eventTimestampDeltaAllowance), startDate as NSDate);
+                            return (events: events, hasMoreEvents: false)
+                        } else if date.timeIntervalSince(timeCursor) > eventTimestampDeltaAllowance {
+                            NSLog("Found event (%@) out of order in history. Ending history fetch.", date as NSDate)
+                            return (events: events, hasMoreEvents: false)
```
ps2: The return here should indicate that the method actually terminated early, since it encountered an anomaly in interpreting pump time.

ps2 (commented):
Thanks for your work on this and the added tests.

In general, I'm finding the test layout a little confusing. Having substantial setup methods (loadSUT, loadTestWithModel, etc) with generic names called from many small tests means I can't read most tests and see what they're testing.  To the extent possible, the test should describe the relevant context in a self contained manner.  If repeating that context setup is too much to bear, then grouping more assertions in a single test is acceptable, or change the name of the setup method to describe the pertinent details.

RileyLinkKit/PumpOpsSynchronous.swift
sharpfive: I agree. Any other way I think to look at the logic seems backwards or confusing too. Any suggestions?

MinimedKit/PumpEvents/PumpEvent.swift
sharpfive: We're encountering a type smell here

PumpModel is required for the initializer as defined in the PumpEvent protocol

Given that, we shouldn't pass it in again on an extension method.

My sense is that a PumpEvent shouldn't care about PumpModel. But some of the PumpEvent subclasses do and that is currently baked into the design.

sharpfive: We should pick one of the cleaner paths

1. Make PumpModel available as part of the PumpEvent interface (or a new construct)
2. Separate PumpModel from PumpEvent

Preference?

RileyLinkKit/PumpOpsSynchronous.swift
sharpfive: Are you thinking to add an extra value to the tuple?

ps2: yes, either an error or a flag indicating loss of interpretability of pump event time.

sharpfive:
Re: the load methods loadSUT and loadTestWithPumpModel

Those are needed because the class under test is constructor injected. We have to run the setUp type code after defining the pumpModel we want to test.

Any ideas for naming them better?

An alternative would be to create a new test class for each various of the dependencies we want to test. This would get unwieldy fast.

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -510,18 +510,24 @@ class PumpOpsSynchronous {
...
-                        NSLog("Found event at (%@) to be more than %@s before startDate(%@)", date as NSDate, String(describing: eventTimestampDeltaAllowance), startDate as NSDate);
-                        return (events: events, hasMoreEvents: false)
-                    } else if date.timeIntervalSince(timeCursor) > eventTimestampDeltaAllowance {
-                        NSLog("Found event (%@) out of order in history. Ending history fetch.", date as NSDate)
-                        return (events: events, hasMoreEvents: false)
-                    } else {
-                        if (date.compare(startDate) != .orderedAscending) {
-                            timeCursor = date
+                    
+                    let possibleCancelBecauseOfDate = !event.canBeDelayedAppend(withPumpModel: pumpModel)
```
ps2: When we hit an event older than startTime, we're not 'canceling' the operation.  We've reached the end successfully.

Also, it's not "can" be delayed, as in a possibility.  When we have the pumpModel and eventType, we know with certainty if it is a delayed append event or not. 

And we can get rid of eventTimestampDeltaAllowance

```
if !event.delayedAppend(withPumpModel: pumpModel) {
    if date <= startDate {
        // Success, we have all the events we need
        //NSLog("Found event at or before startDate(%@)", date as NSDate, String(describing: eventTimestampDeltaAllowance), startDate as NSDate);
        return (events: events, hasMoreEvents: false, cancelledEarly: false)
    } else if date.timeIntervalSince(timeCursor) > TimeInterval(minutes: 60) {
        NSLog("Found event (%@) out of order in history. Ending history fetch.", date as NSDate)
        return (events: events, hasMoreEvents: false, cancelledEarly: true)
    }
}
```

MinimedKit/PumpEvents/PumpEvent.swift
```diff
@@ -21,3 +21,26 @@ public protocol PumpEvent : DictionaryRepresentable {
     }
     
 }
+
+public extension PumpEvent {
+    public func isDelayedAppend(withPumpModel pumpModel: PumpModel) -> Bool {
+        
+        if !pumpModel.mayHaveOutOfOrderEvents {
```
ps2: Boluses (non-square) on both pump models may be out of order, so this method should be removed.

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -510,18 +500,24 @@ class PumpOpsSynchronous {
...
-                    } else if date.timeIntervalSince(timeCursor) > eventTimestampDeltaAllowance {
-                        NSLog("Found event (%@) out of order in history. Ending history fetch.", date as NSDate)
-                        return (events: events, hasMoreEvents: false)
-                    } else {
-                        if (date.compare(startDate) != .orderedAscending) {
-                            timeCursor = date
-                        }
-                        events.insert(TimestampedHistoryEvent(pumpEvent: event, date: date), at: 0)
+                    
+                    let possibleCancelBecauseOfDate = !event.isDelayedAppend(withPumpModel: pumpModel)
```
ps2: We're not 'cancelling' the operation because of date.  When we reach a nonDelayedAppend event that is older than start date, we've finished successfully.  The only time we cancel is when we hit a pump time loss.

RileyLinkKit/PumpOpsSynchronous.swift
```diff
@@ -510,18 +500,24 @@ class PumpOpsSynchronous {
...
+                        if date <= startDate {
+                            // Success, we have all the events we need
+                            //NSLog("Found event at or before startDate(%@)", date as NSDate, String(describing: eventTimestampDeltaAllowance), startDate as NSDate);
+                            return (events: events, hasMoreEvents: false, cancelledEarly: false)
+                        } else if date.timeIntervalSince(timeCursor) > TimeInterval(minutes: 60) {
+                            NSLog("Found event (%@) out of order in history. Ending history fetch.", date as NSDate)
+                            return (events: events, hasMoreEvents: false, cancelledEarly: true)
+                        }                    }
+                    
+                    if (date.compare(startDate) != .orderedAscending) {
```
ps2: With swift 3, dates can be compared using > & <.  I think this can be re-written:

`if date >= startDate`

But I think we just want to update timeCursor for any non-delayed-append events.

MinimedKit/PumpModel.swift
```diff
@@ -50,6 +50,10 @@ public enum PumpModel: String {
         return generation >= 23
     }
     
+    public var mayHaveOutOfOrderEvents: Bool {
```
ps2: This can be removed

MinimedKit/PumpEvents/PumpEvent.swift
```diff
@@ -21,3 +21,22 @@ public protocol PumpEvent : DictionaryRepresentable {
...
+                
+        switch self {
+        case let bolus as BolusNormalPumpEvent:
+            //Square boluses for 523's are appended as the beginning of the event
+            if pumpModel == .Model523 {
+                return bolus.type != .Square
+            }
+            
+            // Square bolus' for some devices are delayed append
+            return bolus.type == .Square
```
ps2: This should just be `return true`.  All pumps have delayed append on normal boluses.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,351 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        let batteryEvent2017 = createBatteryEvent(withDateComponent: dateComponents2007)
+        let pumpEvents: [PumpEvent] = [batteryEvent2017, batteryEvent2007]
+        
+        let (_, _, cancelledEvents) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: Date.distantPast,  pumpModel: pumpModel)
+        
+        XCTAssertTrue(cancelledEvents)
+    }
+    
     func testPumpDateDiscontinuityEventDoesntIncludeEvent() {
         let batteryEvent2007 = createBatteryEvent(withDateComponent: dateComponents2017)
```
ps2: I think these are swapped (2007 != 2017).  Here, and in a few other tests.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,351 @@ class PumpOpsSynchronousTests: XCTestCase {
...
         
         // It should be included
         XCTAssertTrue(array(timeStampedEvents, containsPumpEvent: squareWaveBolus))
     }
+    
+    func testEventAfterDelayedAppendEventFor522IsReturned() {
+        setUpTestWithPumpModel(.Model522)
+        
+        let tempEventBasal = createTempEventBasal2016()
+        let events:[PumpEvent] = [createSquareBolusEvent2010(), tempEventBasal]
```
ps2: The order here is reversed (typically newer events are before older events in the list passed to convertPumpEventToTimestampedEvents()).  This is actually a pump date discontinuity that is being set up, but we're not detecting it because we skip the square wave bolus event.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,351 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        
+        // It should be returned (can't tell if the time for the SquareBolus is valid
+        assertArray(timeStampedEvents, containsPumpEvent: tempEventBasal)
+    }
+    
+    func testEventAfterDelayedAppendEventFor523IsNotReturned() {
+        setUpTestWithPumpModel(.Model523)
+        
+        let tempEventBasal = createTempEventBasal2016()
+        let events:[PumpEvent] = [createSquareBolusEvent2010(), tempEventBasal]
```
ps2: reversed again, but this time it *is* cancelling, and that's the real reason that the tempBasal event isn't included in the results.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
sharpfive: Yes. My typo. Good catch.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
sharpfive: The discontinuity was intentional. I renamed the test to reflect that.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
sharpfive: This test was intentional, the basel event shouldn't be included. I added a new test to verify it was cancelling also.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,351 @@ class PumpOpsSynchronousTests: XCTestCase {
...
         
-        let datePast2007 = dateComponents2007.date!.addingTimeInterval(60*60)
+        let (events, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: datePast2007, pumpModel: pumpModel)
         
-        let batteryEvent2007 = createBatteryEvent(withDateComponent: dateComponents2017)
-        let batteryEvent2017 = createBatteryEvent(withDateComponent: dateComponents2007)
+        XCTAssertEqual(events.count, 1)
+    }
+    
+    func testPumpDateDiscontinuityHasMoreEvents() {
```
ps2: testPumpDateDiscontinuityHasMoreEvents -> testPumpDateDiscontinuityDoesNotHaveMoreEvents

RileyLinkKitTests/PumpOpsSynchronousTests.swift
ps2: This is not how a delayed append event would appear. A delayed append event is going to show up at most 8 hours out of order for a square wave bolus on x22, and at most 15 minutes for boluses on x22.   A pump date discontinuity is a different thing, where a pump loses track of time, and is typically years out of order.

When I read this, it looks like you are testing a pump date loss, and not a delayed append.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,351 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        
+        let tempEventBasal = createTempEventBasal2016()
+        let events:[PumpEvent] = [createSquareBolusEvent2010(), tempEventBasal]
+        let (timeStampedEvents, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: events, startDate: Date.distantPast, pumpModel: pumpModel)
+        
+        // It should not be returned (timestamp from square bolus is valid)
+        assertArray(timeStampedEvents, doesntContainPumpEvent: tempEventBasal)
+    }
 
+    func testDelayedAppendDiscontinuityEventFor522IsReturned() {
```
ps2: Testing what happens in the case of a delayed append *and* a pump date discontinuity (date loss) is quite an edge case.  Because we're skipping delayed append events, we'll miss the pump date discontinuity and return the event, which is maybe a bug, but I'm ok with it, as I don't think it hurts anything, and to fix for it, we'd need to start calculating expected duration of boluses, which a bit of a mess (the pump delivers at different rates based on the size of the bolus, and we haven't fully characterized that behavior into anything concrete that we could use in calculations). 

In short, I think this test is asserting the presence of an issue, and isn't really needed.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,351 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        
+        let squareBolus2016 = createSquareBolusEvent2016()
+        let events:[PumpEvent] = [createSquareBolusEvent2010(), squareBolus2016]
+        let (timeStampedEvents, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: events, startDate: Date.distantPast, pumpModel: pumpModel)
+        
+        //It should be returned
+        XCTAssertTrue(array(timeStampedEvents, containsPumpEvent: squareBolus2016))
+    }
+    
+    func test523SquareBolusEventIsNotReturned() {
```
ps2: This is essentially testing a pump date loss event and that the event signaling the date loss is not included, which is already tested in testPumpDateDiscontinuityEventDoesntIncludeEvent

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,351 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        let squareBolus2016 = createSquareBolusEvent2016()
+        let events:[PumpEvent] = [createSquareBolusEvent2010(), createSquareBolusEvent2010(), squareBolus2016]
+        let (timeStampedEvents, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: events, startDate: Date.distantPast, pumpModel: pumpModel)
+        
+        //It should not be returned
+        XCTAssertFalse(array(timeStampedEvents, containsPumpEvent: squareBolus2016))
+    }
+    
+    func test523RegularBolusDoesReturnEvent() {
+        setUpTestWithPumpModel(.Model523)
```
ps2: In this test, the events look like a pump date discontinuity, but since the event signaling the discontinuity is a delayed append event, we return it.  

This is the same condition and test set up in testDelayedAppendDiscontinuityEventFor522IsReturned, so it should be removed.

ps2 (commented):
I haven't finished going through the tests in PumpOpsSynchronousTests.swift.  I can continue, but maybe you could review the ones starting at test523RegularHasMoreEvents and below in the context of keeping in mind the nuances of the differences between pump date loss and delayed append.  I think some of those tests may be testing things that are already being tested elsewhere.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
         
         let afterBatteryEventDate = batteryEvent.timestamp.date!.addingTimeInterval(TimeInterval(hours: 10))
         
-        let (_, hasMoreEvents) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: afterBatteryEventDate, pumpModel: pumpModel)
+        let (_, hasMoreEvents, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: afterBatteryEventDate, pumpModel: pumpModel)
         
         XCTAssertFalse(hasMoreEvents)
     }
     
     func testBatteryEventIsCreated() {
```
ps2: testEventAfterStartDateIsReturned

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
         let pumpEvents: [PumpEvent] = [createBatteryEvent()]
         
-        let (events, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: Date.distantPast, pumpModel: pumpModel)
+        let (events, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: Date.distantPast, pumpModel: pumpModel)
         
         XCTAssertEqual(events.count, 1)
     }
     
     func testMultipleBatteryEvent() {
+        let batteryEvent2007 = createBatteryEvent(withDateComponent: dateComponents2007)
```
ps2: What behavior is this testing that the previous test doesn't test?  The name should be updated to describe that behavior, or this test removed.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        let batteryEvent2017 = createBatteryEvent(withDateComponent: dateComponents2017)
+        let pumpEvents: [PumpEvent] = [batteryEvent2017, batteryEvent2007]
         
-        let datePast2007 = dateComponents2007.date!.addingTimeInterval(60*60)
+        let (events, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: datePast2007, pumpModel: pumpModel)
         
-        let batteryEvent2007 = createBatteryEvent(withDateComponent: dateComponents2017)
-        let batteryEvent2017 = createBatteryEvent(withDateComponent: dateComponents2007)
+        XCTAssertEqual(events.count, 1)
```
ps2: This should test that the 2007 event is not present in the result.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+    }
+    
+    func testPumpDateDiscontinuityDoesNotHaveMoreEvents() {
```
ps2: Combine the assert from this test and the asserts from the next two tests into the same test, as they are asserting different things about the same condition.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        let (events, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: Date.distantPast, pumpModel: pumpModel)
         
         XCTAssertEqual(events.count, 1)
+        assertArray(events, doesntContainPumpEvent: batteryEvent2017)
     }
     
-    func loadTestWithPumpModel(_ newPumpModel: PumpModel) {
-        pumpModel = newPumpModel
-        loadSUT()
+    func testEventWithSameDataArentAddedTwice() {
```
ps2: testEventsWithSameDataArentAddedTwice

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        let pumpEvents: [PumpEvent] = [createBolusEvent2009(), createBolusEvent2009()]
+        let (events, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: Date.distantPast, pumpModel: pumpModel)
+        XCTAssertEqual(events.count, 1)
     }
     
-    func testMultipleBolusEvents() {
-        loadTestWithPumpModel(.Model522)
+    func testMultipleBolusEventsAreCounted() {
+        setUpTestWithPumpModel(.Model522)
```
ps2: How is this different from tests above that test that events after start date are returned? Are you worried about specific types not being returned. I don't think we want to treat all specific types.

The next two tests are also splitting up the same condition into multiple tests.  If we keep these tests (leaning in favor of not, as I'm not sure they're testing anything new), then combine all three asserts into a single test.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
         
-        let (timeStampedEvents, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: events, startDate: Date.distantPast, pumpModel: pumpModel)
+        let (timeStampedEvents, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: events, startDate: Date.distantPast, pumpModel: pumpModel)
         
         assertArray(timeStampedEvents, containsPumpEvent: createBolusEvent2009())
     }
     
-    func testNonMutableSquareWaveBolusFor522() {
+    func testMultipleBolusEventsWith523ContainsLastEvent() {
+        setUpTestWithPumpModel(.Model523)
```
ps2: Again, this seems to be testing the same logic of including events after start date.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
-
-
-    //shouldFinishIfTimestampBeforeStartDateEncounteredConsideringAdjustedTime()
-    func test523EstimatedTimeDeltaAllowanceBeforeAdjustedStartTime() {
-        let event2010 = createSquareBolusEvent2010()
-        let events = [event2010]
+    
+    // MARK: Discontinuity Event after a Square Bolus
+    func testDiscontinuityEventAfterDelayedAppendEventFor522IsReturned() {
+        setUpTestWithPumpModel(.Model522)
```
ps2: This is testing a corner case that as far as I know, is unspecified behavior.  I actually don't think this is possible to occur in real world situations, and even if it did, I'm not sure what behavior we really want here.  It's misleading to create a test that asserts something we don't actually have an opinion on. 

The same is true of all the following tests down to test523FinalRegularDiscontinuityDoesntCancelsOperation.  They test the situation where a delayed append event is also a pump discontinuity.  They should all be removed.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
-        let batteryEvent2017 = createBatteryEvent(withDateComponent: dateComponents2007)
-        let pumpEvents: [PumpEvent] = [batteryEvent2007, batteryEvent2017]
-        
-        let (events, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: Date.distantPast, pumpModel: pumpModel)
+        let (events, _, _) = sut.convertPumpEventToTimestampedEvents(pumpEvents: pumpEvents, startDate: Date.distantPast, pumpModel: pumpModel)
         
         XCTAssertEqual(events.count, 2)
     }
 
     func testOldBatteryEventIsFiltered() {
```
ps2: testEventBeforeStartDateIsFiltered()

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +82,433 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+        let regularBolusEvent = createBolusEvent2011()
+        let events:[PumpEvent] = [createSquareBolusEvent2010(), createSquareBolusEvent2010(), regularBolusEvent]
+        let (_, _, cancelledEarly) = sut.convertPumpEventToTimestampedEvents(pumpEvents: events, startDate: Date.distantPast, pumpModel: pumpModel)
+        
+        // this doesn't trigger the out of order event cancellation
+        XCTAssertFalse(cancelledEarly)
+    }
+    
+// MARK: Square Bolus Event before starttime (offset 9 minutes)
+    func testSquareBolusEventBeforeStartTimeDoesntContainEvent() {
```
ps2: The fact that sometimes square boluses before start date are returned (523) and sometimes they aren't (522), is a side effect of our deciding to skip delayed append events for time cutoff checking.  It's *not* something that we want to assert, as it is not really desired behavior, but it's acceptable behavior.

ps2 (commented):
more comments in chat...

RileyLinkKitTests/PumpOpsSynchronousTests.swift
sharpfive: Testing multiple events tests the logic around the for loop (or any construct used in the future) to make sure multiple events are included and not skipped somehow. It would be an obvious next test to write in a TDD style workflow. 

You could argue the previous test (testBatteryEventIsCreated) that checks a single value is now redundant and could be deleted. I would keep both tests, but would suggest deleting the first one if you think these are too many.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
sharpfive: I wrote them because they are more specific cases.

I'm OK with deleting them.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -153,87 +145,100 @@ class PumpOpsSynchronousTests: XCTestCase {
...
-        let (timeStampedEvents, hasMoreEvents, cancelled) = sut.convertPumpEventToTimestampedEvents(pumpEvents: events, startDate: Date.distantPast, pumpModel: pumpModel)
-        
-        //It should not be returned
-        XCTAssertFalse(array(timeStampedEvents, containsPumpEvent: squareBolus2016))
-        XCTAssertFalse(hasMoreEvents)
-        XCTAssertTrue(cancelled)
-    }
-    
+
+    // Remove or fixme: on a 523 a square wave bolus followed by a another event will be in correct order. So I'm not sure what this is testing.
```
sharpfive: The PumpEvent array to convertPumpEventToTimestampedEvents should be in descending order.  They are reversed from the order read from the device.

This tests an the same event sequence from the previous test (testDelayedAppendEventDoesNotCauseValidEventsToBeFilteredOut) on a different pump. It explicitly shows a different outcome is expected. In this case the sequence would be flagged as out of order.

I think it's a valuable and instructive test, but not crucial. I'll let you decide if it stays.

Update: the tests had a mismatch of startDate aren't relevant to the tests

RileyLinkKitTests/PumpOpsSynchronousTests.swift
```diff
@@ -73,202 +84,289 @@ class PumpOpsSynchronousTests: XCTestCase {
...
+//        let squareBolusEvent = createSquareBolusEvent2010()
+//        let (timestampedEvents, hasMoreEvents, cancelledEarly) = runDeltaAllowanceTimeTest(pumpEvent: squareBolusEvent, timeIntervalAdjustment: TimeInterval(hours:11))
+//
+//        assertArray(timestampedEvents, doesntContainPumpEvent: squareBolusEvent)
+//        XCTAssertFalse(cancelledEarly)
+//        XCTAssertFalse(hasMoreEvents)
+//    }
+
+
+    // The following tests are testing the border conditionsof the pump lost time detection, the main behavior of which is covered
```
sharpfive, line 302: I'd feel better if we either kept the tests or removed the code they're testing. I don't like leaving in ambiguous code without tests. There may be functionality relying on the current implementation that we're not considering. This type of code can be easily removed by mistake while changing other functionality. 

I'd propose keeping one of these test, probably the one at 59 minutes. One benefit is so we know if the behavior of this method changes.

I'll let you make the call on this. Let me know.

sharpfive (commented):
A couple comments on tests I think are valuable. I'll let you make the call.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
ps2: I can't tell which test you're talking about here... The diff above doesn't give enough context.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
ps2: That sounds fine.

RileyLinkKitTests/PumpOpsSynchronousTests.swift
ps2: My comment on this test is that the situation set up in the test is not what we would expect to see on a 523.  On a 523, the bolus would be put into history at the correct place: before the temp basal in natural history order, and in the order that convertPumpEventToTimestampedEvents expects, after the temp basal.  This is the opposite of what the test sets up.

ps2:
Please continue this work as a new PR against the history-fetch-optimization branch.

PR #336, opened 2017-05-02
Feature/basal schedule command

sharpfive:
Read Basal Schedule Setting from Pump
Also warning cleanup


ps2:
Continuing this work in #337

PR #382, opened 2018-02-25
Fix to NightScout basal rendering requires "percent"

mjliedtke:
A couple of small tweaks to fix NightScout basal rendering.

### DIFF
--- a/NightscoutUploadKit/NightscoutPumpEvents.swift
+++ b/NightscoutUploadKit/NightscoutPumpEvents.swift
@@ -93,6 +93,7 @@ public class NightscoutPumpEvents: NSObject {
             enteredBy: eventSource,
             temp: basalRate.rateType == .Absolute ? .Absolute : .Percentage,
             rate: basalRate.rate,
+            percent: basalRate.rate - 100,
             absolute: absolute,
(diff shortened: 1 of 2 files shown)


NightscoutUploadKit/NightscoutPumpEvents.swift
```diff
@@ -93,6 +93,7 @@ public class NightscoutPumpEvents: NSObject {
             enteredBy: eventSource,
             temp: basalRate.rateType == .Absolute ? .Absolute : .Percentage,
             rate: basalRate.rate,
+            percent: basalRate.rate - 100,
```
ps2, line 96: The RateType enum should be changed to be an enum with associated values.  If the temp basal is a percentage value, then only percent should be represented, and an absolute rate not set, and vice versa.

ps2 (changes requested):

NightscoutUploadKit/NightscoutPumpEvents.swift
mjliedtke: Agreed.  Sorry.  I only use percentage, so wasn't paying attention to what that would do on absolute.

PR #385, opened 2018-03-28
Max settings message clarified

Kdisimone:


MinimedKit/Messages/PumpErrorMessageBody.swift
```diff
@@ -19,7 +19,7 @@ public enum PumpErrorCode: UInt8 {
         case .commandRefused:
             return NSLocalizedString("Command refused; check if pump suspended, prime menu unfinished, or temp basal set to %", comment: "Pump error code returned when command refused")
         case .maxSettingExceeded:
-            return NSLocalizedString("Max setting exceeded", comment: "Pump error code describing max setting exceeded")
+            return NSLocalizedString("Pump setting exceeded; Loop's max basal rate is set higher than the pump's max basal rate", comment: "Pump error code describing max setting exceeded")
```
ps2, line 22: This error also occurs when a bolus larger than the pump's max is attempted.   It probably makes sense to put more details into a recovery suggestion like #386

ps2 (changes requested):

MinimedKit/Messages/PumpErrorMessageBody.swift
ps2: Also, in this context, it's not necessarily Loop, so the message should be changed to be generic.  Loop specific translations/support can be put into Loop.

Kdisimone:
Thanks!  I like the recovery suggestion approach.

PR #402, opened 2018-06-15
Adding new commands for remote control of pump

ps2:
@bewest left us a treasure trove of hints about undecoded minimed commands in a doc file tucked away in the decocare repository: https://github.com/openaps/decocare/blob/master/doc/commands.tsv

There were some gems in there, and now we have commands for:
* Set Basal Schedule (A, B, and STD!)
* Change Max Basal Rate
* Change Max Bolus
* Change Remote Control ID
* Set Remote Control Enabled


Kdisimone:
the gift that keeps giving.  Thanks @bewest

PR #411, opened 2018-06-22
Add Battery, Insulin and Sitechange events.

erikdi:
Use the respective pump events to synthesize these
automatically.  Battery Change will need to be handled
in Loop itself.


ps2:
Thanks @erikdi. Could you test the battery change, and see if you actually get these?  And have you tested the events actually being uploaded to NS, to see how they display?

As for the site change event; assuming that a prime is a site change is obviously not always true.  It would be better if we could upload this as a prime event, and if the user wants to track primes as site changes, that should be a NS option.

erikdi:
I have been running this since ~12 month (https://github.com/erikdi/Loop/blob/dev/Loop/Extensions/NightscoutUploader.swift#L77, https://github.com/erikdi/Loop/blob/dev/Loop/Extensions/DoseStore.swift#L151 ).  The battery change is deducted here from the Loop DeviceDataManager code which is arguably more correct.  These display just fine in Nightscout as IAGE , CAGE and since kdisimone's patch got merged, also as BAGE if the respective plugins are enabled.

The battery event from the pump I just tested and looks like this (It seems one is generated for removal one for adding the battery back):
Event: {
"_type" = Battery;
timestamp = "2018-06-24T17:57:46Z";
}
Event: {
"_type" = Battery;
timestamp = "2018-06-24T17:58:18Z";
}

I'd say actually all 3 events are heuristics, but fit ~95% of the time, saving a bunch of time in tracking diabetes and manual work.  You can still delete these from Nightscout if they are generated in error.

BatteryEvent: Could also be someone just removing and adding the pump.

RewindPumpEvent: Is also generated e.g. after a motor failure when reloading a cartridge, could be the same insulin.

PrimeEvent:  These are generated anytime a RewindPumpEvent is happening as well, even if you keep re-using the same canula in my experiments.  Not sure if there are other ways to generate these without changing the canula?  I don't really think it is worthwhile to track this differently in Nightscout, not sure many users would care about the difference between "Canula Age" and "Time since last Prime".  On the other hand, every diabetes is different so there is probably a usecase I'm overlooking.

Unfortunately it is rather hard to generate these from Loop as it would require a whole new datastore to sync to Nightscout if I don't piggy back on CarbStore or DoseStore.

I could also split the pull request in just adding the Treatment objects (which is hopefully not so controversial) , and a second request for syncing the pump events.

Whatever you prefer.

ps2:
You don't happen to have the history page raw data from the battery event, do you?  I wonder if there is a removed/inserted flag.

ps2:
One use for primes is to make up for an accidental air-bolus; i.e. bolusing accidentally while the site is disconnected.  A prime will let you deliver insulin, without adding IOB to Loop.

ps2:
I would rather that the assumption-making was happening in NS rather than Loop, but this is probably ok.  If we start tracking other events in Loop, we may change this.

ps2:
I need to hold off on it for now, though; we have some pretty big restructuring going on atm.

ps2:
See this: https://github.com/LoopKit/Loop/pull/753/files#diff-61582f29b76c79a09c8db23ee890e2b4L75

And this: https://github.com/LoopKit/Loop/pull/753/files#diff-63dff1398f92f6477343ea302fc1ed5eR13

ps2:
* The insulin change event is good to go. 
* The battery change event needs to incorporate the flag that detects battery present or removed (https://github.com/ps2/rileylink_ios/pull/529), since there are generally a pair, and we don't want two identical "battery replacement events to be logged with each replacement.  
* The assumption that a prime is a site change doesn't work as many people use primes for other reasons, so this needs to be removed or uploaded as a simple prime event.

ps2:
Please submit insulin change as a separate PR. Battery change would also be accepted. But we can't assume prime is site change, as that is often not true, so closing this until either changed as noted, or done via separate PRs.

PR #427, opened 2018-08-11
Adding decoding for unknown event type 57

ps2:


ecc1:
0x39 is the BG Received event (from a meter) for x12/x15 pumps. As far as I can tell it's decoded exactly the same as 0x3F in later pumps.

ps2:
Thanks @ecc1 ! I'll take another look.

PR #428, opened 2018-08-12
Omnikit cancel sounds

Lytrix:
I've updated the tests with cancel temp basal and bolus. 
Separated the bolus tests from the message tests.
Also added the possible sound values to the cancel command which are working great.


Lytrix:
rebased with a new branch, so discarding this one.

PR #429, opened 2018-08-13
Added beepType options to cancel and configure alerts command

Lytrix:
This looks like a lot better PR. ;-)
I've updated cancelDelivery.
Added tests with cancel temp basal and bolus.
Fixed the missing values cancelDelivery variables in the podSessions.

The suspend command still needs to be added into this with this test: 1f 05 b15898b0 `03` which I also have ready, but I wanted to make it testable with a 0x19 command as well.

### DIFF
--- a/OmniKit/MessageBlocks/CancelDeliveryCommand.swift
+++ b/OmniKit/MessageBlocks/CancelDeliveryCommand.swift
@@ -26,17 +26,33 @@ public struct CancelDeliveryCommand : MessageBlock {
         }
     }
     
-    public enum CancelType {
-        case normal
-        case suspend
+    public struct BeepType: OptionSet {
+        public let rawValue: UInt8
         
-        var rawValue: UInt8 {
-            switch self {
-            case .normal:
-                return 0x6
-            case .suspend:
-                return 0x0
-            }
+        static let noBeep = BeepType(rawValue: 0)
+        static let beepBeepBeepBeep = BeepType(rawValue: 1)
+        static let bipBeepBipBeepBipBeepBipBeep = BeepType(rawValue: 2)
+        static let bipBip = BeepType(rawValue: 3)
+        static let beep = BeepType(rawValue: 4)
+        static let beepBeepBeep = BeepType(rawValue: 5)
+        static let beeeeeep = BeepType(rawValue: 6)
+        static let bipBipBipbipBipBip = BeepType(rawValue: 7)
+        static let beeepBeeep = BeepType(rawValue: 8)
+        
+        static let all: BeepType = [
+            .noBeep,
+            .beepBeepBeepBeep,
+            .bipBeepBipBeepBipBeepBipBeep,
+            .bipBip,
+            .beep,
+            .beepBeepBeep,
+            .beeeeeep,
+            .bipBipBipbipBipBip,
+            .beeepBeeep
+        ]
+        
+        public init(rawValue: UInt8) {
+            self.rawValue = rawValue
         }
     }
     
@@ -61,7 +77,7 @@ public struct CancelDeliveryCommand : MessageBlock {
     
     public let deliveryType: DeliveryType
     
-    public let cancelType: CancelType
+    public let beepType: BeepType
     
     let nonce: UInt32
     
@@ -71,7 +87,7 @@ public struct CancelDeliveryCommand : MessageBlock {
             5,
             ])
         data.appendBigEndian(nonce)
-        data.append((cancelType.rawValue << 4) + deliveryType.rawValue)
+        data.append((beepType.rawValue << 4) + deliveryType.rawValue)
         return data
     }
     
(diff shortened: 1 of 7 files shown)


Lytrix:
I've updated this PR to the latest Omnikit branch to review the sound options for cancelDeliveryType. 

It works properly using the Tests commands on a live Pod. I still have some error in Travis to look at, but almost there.

The way I setup the sounds does differ currently with the way it is called on the 0x19 command using `beepType = 0x0302`. 

It is probably better to make one beep method which can be reused in all commands, also to be able to make a setup page to let users choose sound setups for different things. What do you think?

OmniKit/MessageBlocks/CancelDeliveryCommand.swift
```diff
@@ -26,17 +26,33 @@ public struct CancelDeliveryCommand : MessageBlock {
         }
     }
     
-    public enum CancelType {
-        case normal
-        case suspend
+    public struct BeepType: OptionSet {
```
ps2: BeepType should be an enum, not an OptionSet

ps2 (changes requested):
Thanks! Need to get tests passing too.

Lytrix:
Let me know if this is ok. Then I will look at combining the same enum beep setup on the 0x19 command.

ps2 (approved):

OmniKit/MessageBlocks/ConfigureAlertsCommand.swift
```diff
@@ -108,21 +122,24 @@ public struct ConfigureAlertsCommand : MessageBlock {
...
+            
             if encodedData[0] & 0b100 != 0 {
                 let volume = Double(yyyy * 2) * podPulseSize
                 self.expirationType = .reservoir(volume: volume)
             } else {
                 self.expirationType = .time(TimeInterval(minutes: Double(yyyy)))
             }
             
-            self.beepType = (UInt16(encodedData[4]) << 8) + UInt16(encodedData[5])
+            self.beepType = BeepType(rawValue: encodedData[4])!
```
ps2: This should throw an exception if the rawValue isn't valid

OmniKit/MessageBlocks/ConfigureAlertsCommand.swift
```diff
@@ -82,18 +95,19 @@ public struct ConfigureAlertsCommand : MessageBlock {
                 let minutes = UInt16(duration.minutes)
                 data.appendBigEndian(minutes)
             }
-            data.appendBigEndian(beepType)
-            
+            data.appendBigEndian(UInt16(beepType.rawValue) << 8 + UInt16(beepRepeat))
```
ps2: If these are two bytes, you could just append each one in turn, rather than encoding them into a UInt16, and then splitting them back up into something to be appended to the data object.

Lytrix: Good idea, this is a bit overkill.

OmniKit/MessageBlocks/ConfigureAlertsCommand.swift
Lytrix: Do you have an example how I can assert this? Would be good to write a test to see if it fails on values higher than 8.

ps2:
```
guard let beepType = BeepType(rawValue: encodedData[4]) else {
   throw MessageError.unknownValue(value: encodedData[4], typeDescription: "BeepType")
}
self.beepType = beepType
```

OmniKit/MessageBlocks/ConfigureAlertsCommand.swift
```diff
@@ -95,7 +95,8 @@ public struct ConfigureAlertsCommand : MessageBlock {
                 let minutes = UInt16(duration.minutes)
                 data.appendBigEndian(minutes)
             }
-            data.appendBigEndian(UInt16(beepType.rawValue) << 8 + UInt16(beepRepeat))
+            data.append(UInt8(beepType.rawValue))
```
ps2: beepType.rawValue already is a UInt8

OmniKit/MessageBlocks/ConfigureAlertsCommand.swift
```diff
@@ -95,7 +95,8 @@ public struct ConfigureAlertsCommand : MessageBlock {
...
+            data.append(UInt8(beepRepeat))
```
ps2: If you change beepRepeat to a UInt8, then you don't need to worry about casting, and possible errors caused by that (if you cast a number > 255 into a UInt8 you'll get an exception)

Lytrix: Yes, I was unsure where it is essential to explicitly declare the cast type.

OmniKit/MessageBlocks/ConfigureAlertsCommand.swift
```diff
@@ -108,21 +123,27 @@ public struct ConfigureAlertsCommand : MessageBlock {
...
+            
             if encodedData[0] & 0b100 != 0 {
                 let volume = Double(yyyy * 2) * podPulseSize
                 self.expirationType = .reservoir(volume: volume)
             } else {
                 self.expirationType = .time(TimeInterval(minutes: Double(yyyy)))
             }
+            let beepTypeBits = encodedData[4]
+            guard let beepType = BeepType(rawValue: beepTypeBits) else {
+                throw MessageError.unknownValue(value: beepTypeBits, typeDescription: "beepType")
```
ps2: Small suggestion: typeDescription is intended to have name of the enum type that we're missing a value for, so `"beepType"` should be capitalized.

ps2:
Thanks @Lytrix !

PR #454, opened 2018-09-12
add a pink pump color

Kdisimone:
Also, changed the wheel selector text to the full color names; blue, purple, clear, grey, and pink.  Although, now that I'm thinking of it...would probably also be nice to make those localizedStrings since the colors in other languages don't start with the same letters as English nor use same words.  I won't go further with the translation issues though until we know where these strings will end up.


MinimedKit/Models/PumpColor.swift
```diff
@@ -7,8 +7,9 @@
 
 
 public enum PumpColor: String {
-    case blue   = "B"
```
ps2: The intent is for this to be entered based on the letter on the back of the pump. The rest of the design (highlighting the location of those 3 letters visually) and text (describing that the region and color are encoded in those three letters) is built around this.

ps2: I think the letter might have been 'H'?

ps2: And for the image, it looks like the inner sticker part is lighter than the outer plastic casing:
sany0046

ps2 (changes requested):

ps2 (approved):

PR #456, opened 2018-09-15
Omnikit statuserror switcher

Lytrix:
A first shot at a switcher for each statuserror type.

* Added parent type StatusMessageBlock Type to StatusType enum
* Separated Type 5 error command from StatusError
* Made Test passing

Question:
I can get the encodedData[2] in the StatusError struct, but does this pass the data into the StatusMessageBlock switch? If it does, how can I test this?


ps2:
I merged omnikit into dev. You'll need to rebase this branch to dev.

ps2:
If you do a `git diff b376480 > statuserror.diff`, you'll get a diff that can apply to dev HEAD, I think, and you could just start a new branch, if rebasing gets confusing, which I think it might.

Lytrix:
I created a new branch, that was the fastest way. Closing this one.

PR #464, opened 2018-09-19
Dev omnikit podinforesponse

Lytrix:
New PR, did a merge on the dev branch as a new branch.
First tried rebasing/merging the dev on my branch but that was messy. This will ensure the project and plist whitespaces remain untouched.


OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -47,64 +47,58 @@ public struct StatusError : MessageBlock {
...
+    public let data                     : Data
     
     public init(encodedData: Data) throws {
         
-        if encodedData.count < Int(13) {
+        if encodedData.count < Int(16) {
             throw MessageBlockError.notEnoughData
         }
         
         self.length = encodedData[1]
```
ps2: Because this object is now being constructed with `podInfo = try podInfoResponseSubType.podInfoType.init(encodedData: encodedData.subdata(in: 2..<len-2))`, all the array accesses should be shifted minus 2.

ps2 (changes requested):
I think we're close.  Looking forward to getting some of this fault information displayed in the UI

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: Fixed, as well as the tests for these types.

Lytrix:
For the faultEventTypes I tried to make descriptive names, but I can also imagine a simpler case description like faultError92 would also suffice. What do you think is best? I do see some clustering of SetupPod errors and TempBasal errors which could be nice to see back in the code if errors are occuring, so left that way for now:
https://github.com/ps2/rileylink_ios/blob/145dbf7f84263d01606c0b4b8da88cc30e57bdc3/OmniKit/MessageTransport/MessageBlocks/PodInfo.swift#L46

ps2 (approved):

Lytrix:
Moved latest additions into PR on omnipod-testing.

PR #471, opened 2018-09-28
Omnipod testing podinfo

Lytrix:
Latest additions to PodInfo.
- renamed resetStatus>flashVariables
- combined type 50 and 51 to 1 flash decoding struct.


OmniKit/MessageTransport/MessageBlocks/PodInfoFlashLog.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Created by Eelke Jager on 26/09/2018.
+//  Copyright © 2018 Pete Schwamb. All rights reserved.
+//
+
+import Foundation
+
+public struct PodInfoFlashLog : PodInfo {
+    
+    // https://github.com/openaps/openomni/wiki/Command-02-Status-Error-response
+    public var podInfoType     : PodInfoResponseSubType
```
ps2: Please use the existing formatting conventions.  Right vertical alignment that adds padding is useful in situations where it is likely you will want to scan a set of values vertically, looking for gaps, or things like that. Enum values, for instance. Or groups where the item on the right is visually similar to all the other items, with a small variation that is easy to miss, and important to see.  I don't think this falls into either of those categories.

Lytrix:
Merged in branch omnipod-testing.

PR #472, opened 2018-09-28
Get the battery info of rileylink

birdfly:
I found that there is a service （UUID：0x180F）of the Rileylink  ble  protocol which can be read，and it replied its battery level，can we add this info to the Rileylink column？

I always forget to charge my Rileylink，so it will help me a lot if this feature added

image


ps2:
Copying my reply from gitter: The battery service on the ble113 measures the regulated voltage supply, so it will always read the same. That service is more for if you are running the chip off of non lipo batteries
It can’t handle the higher lipo voltage.

PR #473, opened 2018-10-04
Many decoding fixes and added string formatting on PodInfoFaultEvent

Lytrix:
Updated Faultevents from latest message.md from itsmojo.
Fixed decoding errors, like the minutes -> days,  hours, minutes
Created string descriptions for enums to show readable values in the debugger.


OmniKit/Model/FaultEventCode.swift
```diff
@@ -17,12 +17,15 @@ public struct FaultEventCode: CustomStringConvertible {
         case failedFlashErase                     = 0x01
         case failedFlashStore                     = 0x02
         case tableCorruptionBasalExtraCommand     = 0x03
+        case faultEventType04                     = 0x04
```
ps2: FaultEventCode type already handles unknown types, that will stringify as "Fault Event Code %1$02x".  We shouldn't add any enums for types that we don't know anything about.

OmniKit/Model/FaultEventCode.swift
```diff
@@ -120,194 +175,304 @@ public struct FaultEventCode: CustomStringConvertible {
         if let faultType = faultType {
             switch faultType {
             case .noFaults:
-                return LocalizedString("Fault Event Type 0x00: No faults, system running normally.", comment: "Fault Event Type 0x00: No faults, system running normally.")
+                return LocalizedString("Fault Event Type 0x00: No faults, system running normally", comment: "Fault Event Type 0x00: No faults, system running normally")
```
ps2: If you want to prefix *all* faults with "Fault Event Type 0xxx", you can split that out into a prefix generated by the FaultEventCode.description function.

OmniKit/Model/FaultEventCode.swift
```diff
@@ -120,194 +175,304 @@ public struct FaultEventCode: CustomStringConvertible {
...
             case .failedFlashErase:
-                return LocalizedString("Fault Event Type 0x01: Flash erase failed in $4x00 page.", comment: "Fault Event Type 0x01: Flash erase failed in $4x00 page.")
+                return LocalizedString("Fault Event Type 0x01: Flash erase failed in $4x00 page", comment: "Fault Event Type 0x01: Flash erase failed in $4x00 page")
```
ps2: I'm not sure we want the Localize these strings.  Currently they don't surface in Loop, and I'm not sure they should.  Only if the user is going to do different things based on the error.  I think currently all these are debugging/diagnostic level.  I.e. they should be reported, and maybe included in issue reports, but at the top level, I think we tell the user "Pod internal fault xx.  Please generate issue report, deactivate and replace your pod".

OmniKit/Model/FaultEventCode.swift
Lytrix: Good idea, I was also a bit in doubt if this would be a good idea to add. I will remove the unknown ones.

OmniKit/Model/FaultEventCode.swift
Lytrix: The issue reports will be the best place to use them eventually for sure, I will dive into the message types to figure out how to do this properly. Nice thing to look into next week.

OmniKit/Model/FaultEventCode.swift
```diff
@@ -117,197 +117,205 @@ public struct FaultEventCode: CustomStringConvertible {
...
+                return "Fault Event Type 0x8E: Problem inside verify_and_start_pump"
             case .faultEventSetupPodType8F:
-                return LocalizedString("Fault Event Type 0x8F: Bad value during verify_and_start_pump because of sub_ACE3 is stated as bad value.", comment: "Fault Event Type 0x8F: Bad value during verify_and_start_pump because of sub_ACE3 is stated as bad value.")
+                return "Fault Event Type 0x8F: Bad value during verify_and_start_pump because of sub_ACE3 is stated as bad value"
             case .badValueForTables:
-                return LocalizedString("Fault Event Type 0x90: Bad value for $283/$2E3/$315 table specification.", comment: "Fault Event Type 0x90: Bad value for $283/$2E3/$315 table specification.")
+                return "Fault Event Type 0x90: Bad value for $283/$2E3/$315 table specification"
             case .faultEventSetupPodType91:
-                return LocalizedString("Fault Event Type 0x91: Problem inside verify_and_start_pump.", comment: "Fault Event Type 0x91: Problem inside verify_and_start_pump.")
+                return "Problem inside verify_and_start_pump"
```
ps2: The string does not include the fault code, which is important.

OmniKit/Model/FaultEventCode.swift
```diff
@@ -117,197 +117,205 @@ public struct FaultEventCode: CustomStringConvertible {
...
             case .faultEventSetupPodType93:
-                return LocalizedString("Fault Event Type 0x93: Problem inside verify_and_start_pump.", comment: "Fault Event Type 0x93: Problem inside verify_and_start_pump.")
+                return "Problem inside verify_and_start_pump"
             case .badValueField6in0x1A:
-                return LocalizedString("Fault Event Type 0x95: Bad table specifier field6 in 1A command.", comment: "Fault Event Type 0x95: Bad table specifier field6 in 1A command.")
+                return "Bad table specifier field6 in 1A command"
             }
         } else {
-            return String(format: LocalizedString("Fault Event Code %1$02x", comment: "Format string for unknown Fault Event code."), rawValue)
+             return "Unknown Error Code"
```
ps2: This still needs to include the fault number

OmniKit/Model/FaultEventCode.swift
```diff
@@ -117,197 +117,205 @@ public struct FaultEventCode: CustomStringConvertible {
...
         }
     }
 }
+
+extension String {
+    mutating func addFaultCodePrefix(rawValue: UInt8) {
```
ps2: This is not being used.  Also, it can probably just be inline above, in the description var; no need to extend a system class here.

OmniKit/Model/FaultEventCode.swift
```diff
@@ -117,197 +117,199 @@ public struct FaultEventCode: CustomStringConvertible, Equatable {
     }
     
     public var description: String {
+        let prefix = String(format: LocalizedString("Fault Event Code %1$02x:", comment:" Fault Event Code 0xXX"), rawValue)
+
         if let faultType = faultType {
             switch faultType {
             case .noFaults:
-                return LocalizedString("Fault Event Type 0x00: No faults, system running normally.", comment: "Fault Event Type 0x00: No faults, system running normally.")
+                return "\(prefix) No faults, system running normally"
```
ps2:
```
let faultDescription: String
if let faultType = faultType {
    faultDescription = { 
        switch faultType {
        case .noFaults:
            return "No faults, system running normally"
        ....
        }
    }()
} else {
    faultDescription = "Unknown Error Code"
}
return String(format: "Fault Event Code %1$02x: %@", rawValue, faultDescription)
```

Lytrix: I was also looking for a way to return a combined string below, but I did not know where to add the return string variable. So right before the switch, it seems so logical now but I could not find any examples of this. Thanks!

OmniKit/Model/FaultEventCode.swift
Lytrix: Ah yes, that else statement is not in the variable function.

OmniKit/Model/FaultEventCode.swift
```diff
@@ -117,197 +117,202 @@ public struct FaultEventCode: CustomStringConvertible, Equatable {
...
+                    return "Problem inside verify_and_start_pump"
+                case .badValueField6in0x1A:
+                    return "Bad table specifier field6 in 1A command"
+                }
+            }()
         } else {
-            return String(format: LocalizedString("Fault Event Code %1$02x", comment: "Format string for unknown Fault Event code."), rawValue)
+             return "Fault Event Code %1$02x: Unknown Error Code"
         }
+        return String(format: "Fault Event Code %1$02x: %@", rawValue, faultDescription)
```
ps2: The 1$ can be removed

OmniKit/Model/FaultEventCode.swift
```diff
@@ -311,7 +311,7 @@ public struct FaultEventCode: CustomStringConvertible, Equatable {
                 }
             }()
         } else {
-             return "Unknown Error Code"
+            faultDescription = "Fault Event Code %1$02x: Unknown Error Code"
```
ps2: This is not needed... The code is prepended below.

Lytrix: I see it, I am committing too fast, time to go to bed. :-)

OmniKit/Model/FaultEventCode.swift
```diff
@@ -117,197 +117,202 @@ public struct FaultEventCode: CustomStringConvertible, Equatable {
...
+                    return "Problem inside verify_and_start_pump"
+                case .faultEventSetupPodType93:
+                    return "Problem inside verify_and_start_pump"
+                case .badValueField6in0x1A:
+                    return "Bad table specifier field6 in 1A command"
+                }
+            }()
         } else {
-            return String(format: LocalizedString("Fault Event Code %1$02x", comment: "Format string for unknown Fault Event code."), rawValue)
+            faultDescription = "Unknown Error Code"
```
ps2: "Unknown Error Code" -> "Unknown Fault"

MinimedKit/Messages/GetPumpFirmwareVersionMessageBody.swift
```diff
@@ -12,7 +12,7 @@ public class GetPumpFirmwareVersionMessageBody: CarelinkLongMessageBody {
     public let version: String
     
     public required init?(rxData: Data) {
-        let stringEnd = rxData.firstIndex(of: 0) ?? rxData.count
```
ps2: What is this change for?  I don't see an index(of:) function on Data in the docs.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -10,118 +10,133 @@ import Foundation
...
+        self.currentStatus = FaultEventCode(rawValue: encodedData[8])
         
-        self.faultEventTimeSinceActivation = TimeInterval(minutes: Double((Int(encodedData[9] & 0b1) << 8) + Int(encodedData[10])))
+        self.faultEventTimeSinceActivation = TimeInterval(minutes: Double(encodedData[9...10].toBigEndian(UInt16.self)))
         
-        self.insulinRemaining = podPulseSize * Double((Int(encodedData[11] & 0x3) << 8) | Int(encodedData[12]))
+        let resHighBits = Int(encodedData[11] & 0x03) << 6
+        let resLowBits = Int(encodedData[12] >> 2)
+        let reservoirValue = round(Double((resHighBits + resLowBits) * 50)/255)
+        if reservoirValue < StatusResponse.maximumReservoirReading {
```
ps2: This should be set to nil, if we are not at the point where we have a valid reservoir reading.  This is already within the context of having data, so we do not need to distinguish between not being able to get status and this field being set to >50.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -10,118 +10,133 @@ import Foundation
...
         
-        self.originalLoggedFaultEvent = FaultEventCode(rawValue: encodedData[8])
+        self.currentStatus = FaultEventCode(rawValue: encodedData[8])
         
-        self.faultEventTimeSinceActivation = TimeInterval(minutes: Double((Int(encodedData[9] & 0b1) << 8) + Int(encodedData[10])))
+        self.faultEventTimeSinceActivation = TimeInterval(minutes: Double(encodedData[9...10].toBigEndian(UInt16.self)))
         
-        self.insulinRemaining = podPulseSize * Double((Int(encodedData[11] & 0x3) << 8) | Int(encodedData[12]))
+        let resHighBits = Int(encodedData[11] & 0x03) << 6
+        let resLowBits = Int(encodedData[12] >> 2)
```
ps2: Did we determine that these 2 bits do not have valid info?  I meant to look into this.

MinimedKit/Messages/GetPumpFirmwareVersionMessageBody.swift
Lytrix: I had a Travis build error otherwise, it could not find firstIndex and suggested replacing it by index to pass the test. But running it locally it ran ok with firstIndex.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -10,118 +10,133 @@ import Foundation
...
+        self.previousStatus = FaultEventCode(rawValue: encodedData[15])
         
         self.logEventError = encodedData[16] == 2
 
-        self.insulinStateTableCorruption = encodedData[17] & 0b10000000 != 0
-        
-        guard let reservoirStatusAtFirstLoggedFaultEventType = ReservoirStatus(rawValue: encodedData[17] & 0xF) else {
-            throw MessageError.unknownValue(value: encodedData[17] & 0xF, typeDescription: "ReservoirStatus")
+        guard let logEventErrorType = LogEventErrorType(rawValue:encodedData[17] >> 4) else {
+            throw MessageError.unknownValue(value: encodedData[17] >> 4, typeDescription: "LogEventErrorType")
```
ps2: I think this should maybe not throw an exception. I'm guessing we are likely to encounter new values for this, so we should just record it as an unknown type, like we do for FaultCodes.  If we're very confident of not receiving new, unknown values here, then we can throw.

RileyLink.xcodeproj/project.pbxproj
```diff
@@ -2863,6 +2863,7 @@
 					};
 					431CE7761F98564200255374 = {
 						CreatedOnToolsVersion = 9.0;
+						DevelopmentTeam = 86HTWX383F;
```
ps2: Change these back to unset.

ps2 (changes requested):

MinimedKit/Messages/GetPumpFirmwareVersionMessageBody.swift
ps2: Ok, that was probably travis using Xcode 9.  You can change this back; we're running Xcode 10.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: These 2 bits do return the insulinRemaining. I've tested this with a Pod failure where there was 0.05 units not delivered:
`0216020f0000010200ec6a026803ff026b000028a70820230169` 
I will add this as a Unit test.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: Ok, changed this be set to nil.

itsmojo:
To confirm, the entire 10 reservoir field is either $3ff (> 50U) or the number of pulses left (which should only be 0-1000 decimal). I'm not a Swift guy, but I'm not sure why the code is so messy with multiple computations and the rounding. Just extract the 10 bits once into a local variable and if it's not 0x3ff, then set the insulin remaining to be value*podPulseSize. The $3ff is a valid value indicating >= 50 U. Perhaps we might want to differentiate getting the valid > 50 U value and not getting any valid status, but I don't have a strong opinion right now as I'm not familiar with the Swift code implementations.

ps2:
Yes, the comment about looking into those lower bits was that this code had been written before the field was described as well as it is now in the wiki.  Sounds like we should do exactly what you say in terms of extracting the 10 bits as # of pulses, and multiplying by podPulseSize.  As for not getting any valid status, in this context we certainly do have a valid status, so we only need to distinguish between valid data (<= 50) and > 50 (no idea where above 50).

Lytrix:
@itsmojo and @ps2 
Thanks, for the input.
I've cleaned up the code as you stated, but kept in the validation, so anything above 1000 returns nil in this case. I did add a <= so it can start showing 50 if you say that 1000 decimal is also possible.
I will also change it on the StatusResponse part where I got this part of the code from, if this looks ok.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -10,118 +10,106 @@ import Foundation
...
-        
-        guard let reservoirStatusAtFirstLoggedFaultEventCheckType = ReservoirStatus(rawValue: encodedData[19] & 0xF) else {
-            throw MessageError.unknownValue(value: encodedData[19] & 0xF, typeDescription: "ReservoirStatus")
+        guard let previousPodProgressStatus = PodProgressStatus(rawValue: encodedData[19] & 0xF) else {
+            throw MessageError.unknownValue(value: encodedData[19] & 0xF, typeDescription: "PodProgressStatus")
         }
-        self.reservoirStatusAtFirstLoggedFaultEventCheck = reservoirStatusAtFirstLoggedFaultEventCheckType
+        self.previousPodProgressStatus = previousPodProgressStatus
+        
+        self.unKnownValue = encodedData[20...21]
```
ps2: unKnownValue -> unknownValue

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -10,118 +10,106 @@ import Foundation
...
+            "logEventErrorType: \(logEventErrorType.description)",
+            "logEventErrorPodProgressStatus: \(logEventErrorPodProgressStatus)",
+            "recieverLowGain: \(receiverLowGain)",
             "radioRSSI: \(radioRSSI)",
-            "reservoirStatusAtFirstLoggedFaultEventCheck: \(reservoirStatusAtFirstLoggedFaultEventCheck)",
-            "insulinStateTableCorruption: \(insulinStateTableCorruption)",
-            "immediateBolusInProgress: \(immediateBolusInProgress)",
+            "recieverLowGain: \(receiverLowGain)",
+            "previousPodProgressStatus: \(previousPodProgressStatus)",
+            "unKnownValue: \(unKnownValue)",
```
ps2: This is a Data object right? Do you want `unKnownValue.hexadecimalString`?

OmniKit/Model/PodProgressStatus.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  ReservoirStatus.swift
+//  PodState.swift
```
ps2: PodProgressStatus

RileyLink.xcodeproj/project.pbxproj
```diff
@@ -2926,7 +2929,7 @@
 					};
 					C1FFAF7F213323CC00C50C1D = {
 						CreatedOnToolsVersion = 9.4.1;
-						DevelopmentTeam = UY678SP37Q;
+						DevelopmentTeam = 86HTWX383F;
```
ps2: Seems like there are still a few of these?

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: Ah yes, that's what's needed for now.

RileyLink.xcodeproj/project.pbxproj
Lytrix: I removed them from the UI but apparently there was still one left. Thanks for catching that.

ps2: There are a couple more at the bottom of the diff.

PR #476, opened 2018-10-18
Fixed BeepRepeat <-> BeepType order in encoding and decoding.

Lytrix:
@bbassett who is working on the AndroidAPS implementation found a bug which is fixed with this PR:
_In the ConfigureAlerts command, the beep type and beep repeat are listed one way in the document and reversed in the swift code._ 
Fixed the above and also added these:
- Created enum for BeepRepeat to  know what setting you select.
- Fixed Tests to pass on enum values of BeepRepeat
- Removed whitespace from PodInfoTests


Lytrix:
Added these on top of latest omnipod-testing to get a proper rebased version. Will add a new PR.

bbassett:
you've got the wrong guy

PR #477, opened 2018-10-23
Configurealarms beeprepeat fix and added back missing debugDescription values

Lytrix:
@bbassett who is working on the AndroidAPS implementation found a bug which is fixed with this PR:
In the ConfigureAlerts command, the beep type and beep repeat are listed one way in the document and reversed in the swift code.
Fixed the above and also added these:

- Created enum for BeepRepeat to know what setting you select
- Fixed Tests to use enum values of BeepRepeat and fixed the switched values
- Fixed failing PodStateTest


OmniKitTests/PodStateTests.swift
```diff
@@ -13,7 +13,7 @@ class PodStateTests: XCTestCase {
     
     func testNonceValues() {
         
-        var podState = PodState(address: 0x1f000000, activatedAt: Date(), timeZone: .currentFixed, piVersion: "1.1.0", pmVersion: "1.1.0", lot: 42560, tid: 661771)
+        var podState = PodState(address: 0x1f000000, activatedAt: Date(), expiresAt: Date().addingTimeInterval(podSoftExpirationTime), timeZone: .currentFixed, piVersion: "1.1.0", pmVersion: "1.1.0", lot: 42560, tid: 661771)
```
ps2: It doesn't really matter in this context, but might be good to fix nonetheless. `expiresAt` should be calculated using the same date instance that `activatedAt` uses, otherwise they are different times.

ps2: and there are two more below

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -10,10 +10,10 @@ import Foundation
 
 public struct PodInfoFaultEvent : PodInfo, Equatable {
     // https://github.com/openaps/openomni/wiki/Command-02-Status-Error-response
-
+    
     public var podInfoType: PodInfoResponseSubType = .faultEvents
     public let podProgressStatus: PodProgressStatus
-    public let deliveryType: DeliveryType
+    public let deliveryStatus: StatusResponse.DeliveryStatus
```
ps2: I thought DeliveryType was the shared type, so I moved that into Pod.swift.  If DeliveryStatus is the shared type, it should probably move to Pod.swift, and DeliveryType back into the CancelCommand.  Also, DeliveryType doesn't need a description function (the existing one doesn't work for the various combinations of bit fields anyways).

OmniKit/MessageTransport/MessageBlocks/ConfigureAlertsCommand.swift
```diff
@@ -45,6 +45,18 @@ public struct ConfigureAlertsCommand : NonceResyncableMessageBlock {
         case reservoir(volume: Double)
         case time(TimeInterval)
     }
+
+    public enum BeepRepeat: UInt8 {
```
ps2, line 49: Awesome! Love these types. :)

Lytrix: Thanks! Never could have written those so specifically without the hard work of Joe!

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: Yes, I discovered it, when using the test command to ask for a 02 fault event. Then it returned values more consistent with the DeliveryStatus. I switched the two types accordingly and removed the description.

PR #479, opened 2018-10-27
Pairingpod image clip fix

Lytrix:
I accidentally removed the constraints on all items on the pairing screen during my edits in the previous commit. That resulted in the image clipping.

So to be sure nothing else was removed, I redid the  fix applying the margins to the static text' on the previous working storyboard. I've tested this on my iPhone 6 and it looks ok again!

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,1 +1,1 @@
-github "LoopKit/LoopKit" "0f12e7a973625be59ea617166fea7bd2e0f6d729"
+github "LoopKit/LoopKit" "864fc937bd7f30adc1ec3d918a6293987a85d6b1"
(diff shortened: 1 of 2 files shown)


ps2:
There still appear to be quite a few changes to the storyboard file: https://github.com/ps2/rileylink_ios/compare/76a4c95d2c44d3f572ab22a1628dc6839bc2a903...Lytrix:pairingpod-image-clip-fix#diff-bcc9d564cea1e9734987a91d7af88243R1

Lytrix:
Yes, I had to add constraints for multiple screen sizes so that made quite a few changes in the end. It also now stretches the static text box properly when turning to landscape mode.
Also Xcode did some automatic rounding for floats to rounded values for the pixel widths automatically.

ps2:
I expected changes to the the label in question.  There are new constraints on many of the items, changes to multiple controllers, etc.

The changing of the preview from retina3_5 to retina6_1 has changed the static width of many of the entries, which shouldn't affect the final view, but it makes the PR noisy and makes it hard to see what is actually changed. 

Other things I'm wondering about are sectionHeaderHeight changing on RileyLinkSetupTableViewController's table view,  and rowHeight="50" being added. There are also other seemingly random rowHeight changes throughout the diff. 

Not worried about rounding, but the above is a lot more significant.

ps2:
Actually a bigger issue is that we should be using self sizing cells here... I'll take a swing at that.

Lytrix:
I couldn't find a self size option so far on the table cell. Something like fit height view depending on the lines of text, or any scale to fit on the table view, hopefully you can find I way. Maybe there are other layout options which can self size.

ps2:
We already had most of the stuff in place for self sizing. I just needed to change the cells back to default style:  https://github.com/ps2/rileylink_ios/commit/eb3e57c2aff909f743ac31368dc837daafe32a6a

PR #482, opened 2018-11-05
Move suspend/resume button to Pump Settings page

Lytrix:
Moved the suspend/resume button to each of the specific pump settings page of Medtronic and Omnipod. It is now separated from Loop and more logically part of the pump managers.

I thought of moving the different parts to common folder into one suspend/resume file and reference elements from there. But then the function does get a bit more obscured that way so I did not do this.

Suspend/resume could also become part of the configuration, but because it is such a specific failsafe/shower button, a separate item feels more suited then a config list item.

### DIFF
--- a/MinimedKitUI/MinimedPumpSettingsViewController.swift
+++ b/MinimedKitUI/MinimedPumpSettingsViewController.swift
@@ -34,7 +34,8 @@ class MinimedPumpSettingsViewController: RileyLinkSettingsViewController {
 
         tableView.sectionHeaderHeight = UITableView.automaticDimension
         tableView.estimatedSectionHeaderHeight = 55
-
+        
+        tableView.register(SuspendResumeTableViewCell.self, forCellReuseIdentifier: SuspendResumeTableViewCell.className)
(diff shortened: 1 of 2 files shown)


MinimedKitUI/MinimedPumpSettingsViewController.swift
```diff
@@ -34,7 +34,8 @@ class MinimedPumpSettingsViewController: RileyLinkSettingsViewController {
 
         tableView.sectionHeaderHeight = UITableView.automaticDimension
         tableView.estimatedSectionHeaderHeight = 55
-
+        
+        tableView.register(SuspendResumeTableViewCell.self, forCellReuseIdentifier: SuspendResumeTableViewCell.className)
```
ps2, line 38: Omnipod needs to register this cell as well.

Lytrix: Ah yes, forgot that one. Thanks!

ps2:
Resolved by https://github.com/ps2/rileylink_ios/commit/25408269b71342f65543a3e127af1f19422d8b00

PR #486, opened 2018-12-11
adding prompt to check temp basal type

Kdisimone:
Took a wild stab at it.  Storyboards aren't my thing, but I tried.


ps2:
See the clock setup view controller for an example of a screen that can be inserted in the flow as an important piece of information.

PR #488, opened 2019-01-09
Fault Event and Issue Report updates

Lytrix:
- Updated/Added new fault events up to commit 8bab297 of events.md
- Updated IssueReport Faultevents with hex and total insulin delivered values
- Updated PodInfoTest for new values


OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -96,27 +97,28 @@ public struct PodInfoFaultEvent : PodInfo, Equatable {
...
         return [
             "## PodInfoFaultEvent",
             "* rawHex: \(data.hexadecimalString)",
             "* podProgressStatus: \(podProgressStatus)",
             "* deliveryStatus: \(deliveryStatus.description)",
-            "* insulinNotDelivered: \(insulinNotDelivered) U",
-            "* podMessageCounter: \(podMessageCounter)",
-            "* unknownPageCode: \(unknownPageCode.hexadecimalString)",
+            "* insulinNotDelivered: (%.2f, \(insulinNotDelivered)) U",
+            "* podMessageCounter: 0x\(podMessageCounter)",
```
ps2: The 0x prefix implies, hex, but this is rendering decimal

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -16,13 +16,13 @@ public struct PodInfoFaultEvent : PodInfo, Equatable {
     public let deliveryStatus: DeliveryStatus
     public let insulinNotDelivered: Double
     public let podMessageCounter: UInt8
-    public let unknownPageCode: Data
-    public let previousStatus: FaultEventCode
+    public let totalInsulinDelivered: Double
     public let currentStatus: FaultEventCode
     public let faultEventTimeSinceActivation: TimeInterval?
-    public let reservoirLevel: Double?
+    public let reservoirLevel: String
```
ps2: The internal rep of this should stay `Double?`.  Conversion to string form should be done at display time.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -96,27 +97,28 @@ public struct PodInfoFaultEvent : PodInfo, Equatable {
...
     public var debugDescription: String {
         return [
             "## PodInfoFaultEvent",
             "* rawHex: \(data.hexadecimalString)",
             "* podProgressStatus: \(podProgressStatus)",
             "* deliveryStatus: \(deliveryStatus.description)",
-            "* insulinNotDelivered: \(insulinNotDelivered) U",
-            "* podMessageCounter: \(podMessageCounter)",
-            "* unknownPageCode: \(unknownPageCode.hexadecimalString)",
+            "* insulinNotDelivered: (%.2f, \(insulinNotDelivered)) U",
```
ps2: I don't think this is doing what you are intending. It will render like `* totalInsulinDelivered: (%.2f, 0.0) U`

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -61,16 +62,16 @@ public struct PodInfoFaultEvent : PodInfo, Equatable {
...
+            self.reservoirLevel = String(format:"%.2f", reservoirValue)
         } else {
-            self.reservoirLevel = nil
+            self.reservoirLevel = "50+"
         }
         
         self.timeActive = TimeInterval(minutes: Double(encodedData[13...14].toBigEndian(UInt16.self)))
         
-        self.previousStatus = FaultEventCode(rawValue: encodedData[15])
+        self.unacknowledgedAlerts = Data(encodedData[15])
```
ps2: If this is a single value, then it should be an integer type.

ps2 (changes requested):

Lytrix:
Thanks for the fast reviewing @ps2 ! I've updated the parts that were wrong. 
I only haven't been able to test if reservoirLevel renders default as Optional("50+") or as "50+".

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: fixed

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: reverted back to Double? and formatted at display time.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: kept decimal to render, only used it for unknownValue

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
Lytrix: made an extension for this.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
```diff
@@ -63,14 +64,14 @@ public struct PodInfoFaultEvent : PodInfo, Equatable {
...
             self.reservoirLevel = reservoirValue
         } else {
-            self.reservoirLevel = nil
+            self.reservoirLevel =  nil
         }
         
         self.timeActive = TimeInterval(minutes: Double(encodedData[13...14].toBigEndian(UInt16.self)))
         
-        self.previousStatus = FaultEventCode(rawValue: encodedData[15])
+        self.unacknowledgedAlerts = encodedData[15]
```
ps2, line 72: Does this field map to AlertType?

OmniKitTests/PodInfoTests.swift
```diff
@@ -225,15 +225,15 @@ class PodInfoTests: XCTestCase {
             XCTAssertEqual(.faultEvents, decoded.podInfoType)
             XCTAssertEqual(.errorEventLoggedShuttingDown, decoded.podProgressStatus)
             XCTAssertEqual(.suspended, decoded.deliveryStatus)
-            XCTAssertEqual(0000, decoded.insulinNotDelivered)
+            XCTAssertEqual(0, decoded.insulinNotDelivered)
             XCTAssertEqual(4, decoded.podMessageCounter)
-            XCTAssertEqual("07eb", decoded.unknownPageCode.hexadecimalString)
+            XCTAssertEqual("50.15", decoded.totalInsulinDelivered.twoDecimals)
```
ps2, line 230: Probably better to test actual value here, than string representation.  The .twoDecimals method is a private extension for formatting the debugDescription, not for external use.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
ps2: @itsmojo This is the TT field.  Should it map to these types?https://github.com/ps2/rileylink_ios/blob/f0191d2805d782a920f16b816d20c69f9c2e8b07/OmniKit/MessageTransport/MessageBlocks/ConfigureAlertsCommand.swift#L34-L42

itsmojo: Yes, as long as Loop continues to the same conventions for the alerts that the PDM uses.

itsmojo:
For the type2 fault info, the insulinNotDelivered and totalInsulinDelivered fields don't need and should not have any bit masking as they are both returned as 16-bit word values. This is unlike the $1D response where these same values are squeezed into 11-bit and 13-bit fields for minimal message size (heck, they even dropped the message length byte for the $1D response to make it smaller). And in any case, the unneeded insulinNotDelivered bit masking for the type2 fault info was incorrect as this is actually an 11-bit value.

OmniKit/MessageTransport/MessageBlocks/PodInfoFaultEvent.swift
ps2: In the [$19 Configure Alerts](https://github.com/openaps/openomni/wiki/Command-19-Configure-Alerts) command, the alarm type is not a bitmask, but a single value, which we decode to the `AlertType` shown above.  The [wiki for this TT field](https://github.com/openaps/openomni/wiki/Command-02-Pod-Information-Response#type-2) indicates it's a bitmask, so are the `AlertType` values translated into bit positions and used as a bitmask, or is this field the same as the `aaaaaaaa` bits in the $1d response?  The latter is decoded into PodAlarm (https://github.com/ps2/rileylink_ios/blob/bea1f163bc98f092ef9e6ee9e42480fb14377635/OmniKit/Model/PodAlarm.swift#L11-L19)

ps2: Sounds like we need to do a bigger refactor here; PodAlarm and AlertTypes should be based off of the same values; we can take that up in a different PR.

PR #501, opened 2019-03-08
Add new pump precision methods required by PumpManager protocol

ps2:


MinimedKit/PumpManager/MinimedPumpManager.swift
```diff
@@ -16,19 +16,38 @@ public protocol MinimedPumpManagerStateObserver: class {
...
+
+    public func roundToSupportedBolusVolume(units: Double) -> Double {
+        return supportedBolusVolumes.filter({$0 <= units}).max() ?? 0
+    }
 
     public var supportedBasalRates: [Double] {
         return state.pumpModel.supportedBasalRates
     }
 
+    public var supportedBolusVolumes: [Double] {
```
darinkrauss, line 31: [nit] Order here the same as in the protocol?

MinimedKit/Models/PumpModel.swift
```diff
@@ -115,6 +115,18 @@ public enum PumpModel: String {
...
+    public var maximumBolusVolume: Double {
+        return 25
+    }
+
+    public var maximumBasalRate: Double {
+        return 35
+    }
+
+    public var supportedBolusVolumes: [Double] {
+        return supportedBasalRates.filter { $0 < maximumBolusVolume }
```
darinkrauss: I think `<` should be `<=`.

MinimedKit/Models/PumpModel.swift
```diff
@@ -115,6 +115,18 @@ public enum PumpModel: String {
         }
     }
 
+    public var maximumBolusVolume: Double {
+        return 25
+    }
+
+    public var maximumBasalRate: Double {
+        return 35
```
darinkrauss, line 123: Really!?! Wow. (Nothing wrong, just shocked this is so high.)

darinkrauss (changes requested):
Only one thing that really needs a look at, plus a nit for your consideration.

darinkrauss (approved):
LGTM 🚀

PR #503, opened 2019-03-21
Bolus progress and cancelling support

ps2:


MinimedKit/Models/PumpModel.swift
```diff
@@ -134,6 +127,27 @@ public enum PumpModel: String {
...
+    }
+
+    public func bolusDeliveryTime(units: Double) -> TimeInterval {
+        let unitsPerMinute: Double
+        if variableDeliveryRate {
+            switch units {
+            case let u where u < 1.0:
+                unitsPerMinute = 0.75
+            case let u where u > 7.5:
+                unitsPerMinute = units / 5
```
darinkrauss, line 142: Is this really `units / 5`? Meaning that if I'm bolusing 7.7 it will deliver 1.54 units per minute? Just curious.

MinimedKit/Models/PumpModel.swift
```diff
@@ -134,6 +127,27 @@ public enum PumpModel: String {
     public var minimumBasalScheduleEntryDuration: TimeInterval {
         return .minutes(30)
     }
+
+    public var variableDeliveryRate: Bool {
```
mpangburn: Nit: more clearly a boolean as `isDeliveryRateVariable`

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
```diff
@@ -0,0 +1,125 @@
...
+                self.notify()
+            }
+        }
+        runLoop.add(timer, forMode: .default)
+        self.timer = timer
+    }
+
+    private func stop() {
+        timer?.invalidate()
+    }
```
darinkrauss, line 77: From MockDoseProgressEstimator: `timer = nil`?

darinkrauss: at the end of `stop`.

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
```diff
@@ -0,0 +1,125 @@
...
+}
```
darinkrauss, line 78: From MockDoseProgressEstimator:  `deinit`?

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
```diff
@@ -0,0 +1,125 @@
...
+    init(dose: DoseEntry, pumpModel: PumpModel) {
+        self.dose = dose
+        self.pumpModel = pumpModel
+    }
+
+    func addObserver(_ observer: DoseProgressObserver) {
+        os_unfair_lock_lock(&lock)
+        defer {
+            os_unfair_lock_unlock(&lock)
+        }
```
mpangburn: Note re os_unfair_lock in LoopKit PR applies here also.

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
```diff
@@ -0,0 +1,125 @@
...
+//  MinimedKit
+//
+//  Created by Pete Schwamb on 3/14/19.
+//  Copyright © 2019 Pete Schwamb. All rights reserved.
+//
+
+import Foundation
+import LoopKit
+
+class MinimedDoseProgressEstimator: DoseProgressEstimator {
```
darinkrauss: Overall, might want to make MockDoseProgressEstimator match much of MinimedDoseProgressEstimator. In fact, they are similar enough that one might imagine a base class implementation. Just a thought.

MinimedKit/PumpManager/MinimedPumpManager.swift
```diff
@@ -205,17 +192,36 @@ public class MinimedPumpManager: RileyLinkPumpManager, PumpManager {
     }
     
     // MARK: - PumpManager
+    private enum SuspendTransition {
+        case suspending
+        case resuming
+    }
+
+    // TODO: Accessed and set on different threads
```
mpangburn, line 205: Any reason to hold up this TODO? (I assume the action item is synchronizing access to `suspendTransition`)

mpangburn (approved):
🚀

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
```diff
@@ -0,0 +1,125 @@
...
+                updateResolution = 40 // Resolution = 0.025
+                unroundedVolume = timeProgress * dose.units
+            } else {
+                var remainingUnits = dose.units
+                var baseDuration: TimeInterval = 0
+                var overlay1Duration: TimeInterval = 0
+                var overlay2Duration: TimeInterval = 0
+                let baseDeliveryRate = 1.5 / TimeInterval(minutes: 1)
+
+                baseDuration = min(duration, remainingUnits / baseDeliveryRate)
```
darinkrauss, line 37: I honestly don't understand this logic for lines 43-52. 

Assuming `duration` (`endDate` - `startDate`) will always be greater than the remaining time left (`remainingUnits / baseDeliveryRate`) then `baseDuration = remainingUnits / baseDeliveryRate`. Substituting into the second line `remainingUnits -= (remainingUnits / baseDeliveryRate) * baseDeliveryRate` which turns into `remainingUnits -= remainingUnits` which is 0.  Thus, `overlay1Duration` and `overlay2Duration` will always be 0, as well.

But, then again, I don't get what you are trying to do here either, so maybe I'm just missing something. :)

darinkrauss: Or does `remainingUnits` start out as the full dose? Perhaps I don't understand what `endDate` represents.

Yea, I'm confused (which says more about my lack of knowledge than anything else).

MinimedKit/PumpManager/MinimedPumpManager.swift
```diff
@@ -659,29 +669,48 @@ public class MinimedPumpManager: RileyLinkPumpManager, PumpManager {
...
                 completion(.failure(error))
             }
         }
     }
 
+    public func cancelBolus(completion: @escaping (PumpManagerResult<DoseEntry?>) -> Void) {
+        self.bolusState = .canceling
+        setSuspendResumeState(state: .suspend) { (error) in
+            if let error = error {
+                completion(.failure(error))
```
darinkrauss, line 707: Do we want to leave `bolusState = .canceling` if there is an error?

darinkrauss: [BTW, I'm just cheating by comparing the implementations of Mock* and Minimed*.]

MinimedKit/PumpManager/MinimedPumpManager.swift
```diff
@@ -659,29 +669,48 @@ public class MinimedPumpManager: RileyLinkPumpManager, PumpManager {
...
+                let deliveryTime = self.state.pumpModel.bolusDeliveryTime(units: units)
                 let endDate = date.addingTimeInterval(deliveryTime)
                 let dose = DoseEntry(type: .bolus, startDate: date, endDate: endDate, value: units, unit: .units)
                 willRequest(dose)
 
                 try session.setNormalBolus(units: units)
-                completion(.success(dose))
+
+                // Between bluetooth and the radio and firmware, about 2s on average passes before we start tracking
+                let commsOffset = TimeInterval(seconds: -2)
```
darinkrauss, line 687: So, you mean the bolus actually started two seconds ago (on average)?

darinkrauss (changes requested):
A few possible things to look into. Primarily, differences between the Mock* and Minimed* versions of these types, and some logic I don't understand (which could just be my lack of understanding).

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
darinkrauss: Looks like line #s changed to 37-46.

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
darinkrauss: Resolved.

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
darinkrauss: Resolved.

MinimedKit/Models/PumpModel.swift
ps2: Yes, on average, over that 5 minutes. In reality there are three different rates, each pulsed once every 2 seconds. Fastest at the beginning, then slowing down, all adding up to the amount of delivery over 5 minutes.

MinimedKit/PumpManager/MinimedPumpManager.swift
ps2: This is coming in a separate PR.

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
ps2: The base delivery rate is 1.5 U/m, which is a 0.05 pulse every 2s.  If, at this rate, it cannot deliver the entire dose within 5 minutes, it starts double pulsing (0.1U), but keeping the same periodicity (every 2s). It double pulses for just long enough that the remainder will finish out in 5 minutes at the lower 1.5U/m, single pulse rate.  If the dose is large enough that it can't deliver in 5 minutes even by double pulsing, it will start triple pulsing.  I didn't know this was precisely how it did it until last week, when I had to figure it out. :)

MinimedKit/PumpManager/MinimedPumpManager.swift
ps2: Yes

darinkrauss (changes requested):
There are some comments that have not been addressed, I think.

MinimedKit/PumpManager/MinimedDoseProgressEstimator.swift
darinkrauss: Thanks for the clarification. So, `duration` is capped at a maximum of 5 minutes elsewhere?

ps2: Duration is calculated by PumpModel.bolusDeliveryTime, and for x23 pumps, it will be a max of 5 minutes

darinkrauss: :+1:

darinkrauss (approved):
LGTM 🚢

PR #505, opened 2019-03-28
Fix framework search path for RileyLinkBLEKit target

darinkrauss:
@ps2


ps2:
Thanks!  I think this fixes an issue I was hitting yesterday.

PR #506, opened 2019-04-02
Add option for acknowledgementBeep

itsmojo:


OmniKit/MessageTransport/MessageBlocks/BasalScheduleExtraCommand.swift
```diff
@@ -12,6 +12,7 @@ public struct BasalScheduleExtraCommand : MessageBlock {
 
     public let blockType: MessageBlockType = .basalScheduleExtra
     
+    public let acknowledgeReminder: Bool
```
ps2: audibleConfirmation is probably a better name for this variable.

ps2:
Looks like tests need to be updated with the new method signatures.

OmniKit/MessageTransport/MessageBlocks/BasalScheduleExtraCommand.swift
itsmojo: I think that calling this audibleConfirmation would be confusing as this is just one of 3 different types of audible confirmations for the pod when programming insulin commands. The PDM calls the other two types of audio confirmations the Confidence Reminder and the Program Reminder. All of 3 of these different reminders (audio confirmations) are encoded into a single byte referred to in the wiki and in Loop app as the reminder byte. The PDM doesn’t apparently use nor have a name for this particular type of reminder (basically it's the first half of a Confidence Reminder which only beeps when the command with the $80 bit set in the Reminder byte is accepted), so I have been calling this bit an Acknowledge Reminder in the wiki and now in the Loop app to fit with the PDM naming scheme.

itsmojo:
Updated the tests for the new signature methods

OmniKit/MessageTransport/MessageBlocks/BasalScheduleExtraCommand.swift
ps2: It's just not a "reminder". I was confused about the purpose of this flag when I saw it, due to that. Perhaps audibleAcknowledge? audibleCommandReceiptConfirmation?

PR #514, opened 2019-04-07
On pod fault display Pod expired/Empty reservoir/Occlusion detected/fault code

itsmojo:


OmniKitUI/ViewControllers/ReplacePodViewController.swift
```diff
@@ -20,19 +20,36 @@ class ReplacePodViewController: SetupTableViewController {
         case canceledPairing
     }
 
+    private var faultType: FaultEventCode.FaultEventType?
```
ps2: It'd be a little cleaner to make the PodReplacementReason.fault case contain the fault type.

OmniKitUI/ViewControllers/ReplacePodViewController.swift
```diff
@@ -15,37 +15,50 @@ class ReplacePodViewController: SetupTableViewController {
 
     enum PodReplacementReason {
         case normal
-        case fault
+        case fault(_ eventType: FaultEventCode.FaultEventType)
```
ps2: If you make the fault case take a PodInfoFaultEvent, then you can handle the unknown types as well.

OmniKitUI/ViewControllers/ReplacePodViewController.swift
```diff
@@ -15,37 +15,50 @@ class ReplacePodViewController: SetupTableViewController {
...
         didSet {
             pumpManager.getPodState { (podState) in
                 DispatchQueue.main.async {
-                    if podState?.fault != nil {
-                        self.replacementReason = .fault
+                    if let podFault = podState?.fault {
+                        if let faultType = podFault.currentStatus.faultType {
+                            self.replacementReason = .fault(faultType)
+                        } else {
+                            self.replacementReason = .fault(.valuesDoNotMatchOrAreGreaterThan0x97)
```
ps2: Setting a .valuesDoNotMatchOrAreGreaterThan0x97 type when we don't have a faultType means that the UI will show 152 (0x98) for unknown fault codes, like if we happen to get an 0x04 back from the pod for some reason. If you have the .fault case store a PodInfoFaultEvent or FaultEventCode, then you'll have access to the actual fault codes for these.

OmniKitUI/ViewControllers/ReplacePodViewController.swift
```diff
@@ -15,37 +15,50 @@ class ReplacePodViewController: SetupTableViewController {
...
+                    instructionsLabel.text = LocalizedString("Pod expired. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Instructions when replacing pod due to an expired pod")
+                case .occluded,                    // $14
+                     .occlusionCheckValueTooHigh, .occlusionCheckStartup1, .occlusionCheckStartup2,         // $5A, $60, $60
+                     .occlusionCheckTimeouts1, .occlusionCheckTimeouts2, .occlusionCheckTimeouts3,          // $62, $66, $67
+                     .occlusionCheckPulseIssue, .occlusionCheckBolusProblem, .occlusionCheckAboveThreshold: // $68, $69, $6A
+                    instructionsLabel.text = LocalizedString("Occlusion detected. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Instructions when replacing pod due to an occlusion")
+                default:
+                    instructionsLabel.text = String(format: LocalizedString("The pod has detected internal fault 0x%02x. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Instructions when replacing pod due to a fault (1: The fault code value)"), eventType.rawValue)
```
ps2: I wonder if we should show the decimal fault code here, as that's what the pdm does.  Even though we have been tending to use the hex code in our own discussions a lot.

ps2 (changes requested):

OmniKitUI/ViewControllers/ReplacePodViewController.swift
itsmojo: I wondered about this as well, but decided to stick with using a hex numbers after talking to Katie DiSimone who suggesting doing this as we'd only be documenting to call Insulet for something like an obvious pod failure at startup and not for any screamers and so it would be better to keep our fault info more separated. The PDM errors report #'s have a lot more information and have a different format (XX-XXXXX-XXXXX-XXX) with the last 3 digits being the fault # is decimal.

OmniKitUI/ViewControllers/ReplacePodViewController.swift
itsmojo: I thought that mapping unknown types to our code to a special fault code which indicates that the fault code didn't match anything would have been the preferred way to go on something like this, but I would have rather just printed the actual fault code. Sigh. Will try to rework again...

ps2: This isn't a question of preference; it wasn't doing what you were wanting it to do (i.e. display the actual fault code).  It was displaying 0x98 for unknown codes that weren't 0x98.

itsmojo: Yes, I understood this that would have been the case. In any case, no longer relevant with PodReplacementReason.fault now having the actual fault information.

OmniKitUI/ViewControllers/ReplacePodViewController.swift
```diff
@@ -25,19 +25,23 @@ class ReplacePodViewController: SetupTableViewController {
...
+                         .occlusionCheckTimeouts1, .occlusionCheckTimeouts2, .occlusionCheckTimeouts3,          // $62, $66, $67
+                         .occlusionCheckPulseIssue, .occlusionCheckBolusProblem, .occlusionCheckAboveThreshold: // $68, $69, $6A
+                        instructionsLabel.text = LocalizedString("Occlusion detected. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Instructions when replacing pod due to an occlusion")
+                    default:
+                        instructionsLabel.text = String(format: LocalizedString("The pod has detected internal fault %d. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Instructions when replacing pod due to a fault (1: The fault code value)"), faultType.rawValue)
+                    }
+                } else {
+                    instructionsLabel.text = LocalizedString("The pod has detected an internal fault. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Instructions when replacing pod due to a fault ")
```
ps2: You can use `faultCode.rawValue` here to show the actual fault code number.

itsmojo: If I try to use faultCode.rawValue instead of faultType.rawValue to show the actual fault code number on line 41, I get a Swift Compile Error of "‘rawValue’ is inaccessible due to ‘internal’ protection level". The current code compiles cleanly and handles all the simulated fault conditions I injected correctly.

ps2: Ah, yes, you'll need to expose rawType on FaultEventCode as public. This is the only way to access codes like 0x04, or 0x0C if we ever happened to receive them from a pod.

itsmojo: Yes. This has been the underlying issue all along and why I originally was mapping unknown fault errors to a DoNotMatch error and then to a generic internal fault error with no #. So do you now want me to expose rawType on FaultEventCode and rework the ReplacePodViewController code to always display the pod reported fault # independent of whether it maps to something or not? Does the code displaying faults for an Issue Report always report the pod fault # right now and if not, should this get stuff reworked as well once the the rawType on FaultEventCode is made public?

OmniKitUI/ViewControllers/ReplacePodViewController.swift
```diff
@@ -26,27 +26,20 @@ class ReplacePodViewController: SetupTableViewController {
...
-                    case .reservoirEmpty:              // $18
-                       instructionsLabel.text = LocalizedString("Empty reservoir. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Instructions when replacing pod due to an empty reservoir")
-                    case .exceededMaximumPodLife80Hrs: // $1C
-                       instructionsLabel.text = LocalizedString("Pod expired. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Instructions when replacing pod due to an expired pod")
-                    case .occluded,                    // $14
-                         .occlusionCheckValueTooHigh, .occlusionCheckStartup1, .occlusionCheckStartup2,         // $5A, $60, $60
-                         .occlusionCheckTimeouts1, .occlusionCheckTimeouts2, .occlusionCheckTimeouts3,          // $62, $66, $67
-                         .occlusionCheckPulseIssue, .occlusionCheckBolusProblem, .occlusionCheckAboveThreshold: // $68, $69, $6A
+                switch faultCode.rawValue {
+                case 0x18:
```
ps2: We shouldn't use raw codes when we have a decoded enum version of them.  The logic for determining a localized description of a faultcode probably belongs on the FaultEventCode struct itself:

```
    public var localizedDescription: String {
        if let faultType = self.faultType {
            switch faultType {
            case .reservoirEmpty:
                return LocalizedString("Empty reservoir", comment: "Description for empty reservoir fault")
            case .exceededMaximumPodLife80Hrs:
                return LocalizedString("Pod expired", comment: "Description for pod expired fault")
            case .occluded,
                .occlusionCheckValueTooHigh, .occlusionCheckStartup1, .occlusionCheckStartup2,
                .occlusionCheckTimeouts1, .occlusionCheckTimeouts2, .occlusionCheckTimeouts3,
                .occlusionCheckPulseIssue, .occlusionCheckBolusProblem, .occlusionCheckAboveThreshold:
                return LocalizedString("Occlusion detected", comment: "Description for occlusion detected fault")
            default:
                return String(format: NSLocalizedString("The pod has detected internal fault. Code: %1$d", comment: "Format string for undescribed faults (1: The fault code value)"), rawValue)
            }
        } else {
            return String(format: LocalizedString("Unknown pod fault %1$d", comment: "Format string for unknown faults (1: The fault code value)"), rawValue)
        }
    }
```

Then, the code here becomes: 
```
instructionsLabel.text = String(format: LocalizedString("%1$@. Insulin delivery has stopped. Please deactivate and remove pod.", comment: "Format string providing instructions for replacing pod due to a fault. (1: The fault description)"), faultCode.localizedDescription)
```

itsmojo: + I thought we wanted to display the actual pod fault # in decimal in the case of a fault that doesn't match anything and for matching fault #'s other than the Pod expired/Reservoir empty/Occlusion detected cases for which we have simplified messages which match the PDM (i.e., we don't want to display our internal fault code descriptions to the user in the replace pod screens, just in the Loop Report). So we really don't actually want to use any faultCode decoded strings here, do we?
+ The above example is based on the previous PR which used the .reservoirEmpty style case names which doesn't currently support printing the pod fault # for codes we don't have a mapping for. In the latest PR, FaultEventCode's rawValue is public as per the earlier recommendation so that all fault #'s can easily be reported as returned from the pod which is what I've always wanted to do and which you indicated you wanted as well. Do you want to abandon this approach and do something different now?
+ In any case, using the suggested code the build fails with "Value of type 'FaultEventCode' has no member 'localizedDescription'"

ps2: The `localizedDescription` code I showed above would be implemented on FaultEventCode.

itsmojo: Reworked to use a new faultCode.localizedDescription which returns a localized version of "Unknown pod fault %03d" (for a fault that Loop doesn't have an enum case for), "Occlusion detected", "Empty reservoir", "Pod expired",  or "Internal pod fault %03d" (for any other fault).

ps2 (approved):

PR #531, opened 2019-07-28
OverrideStatus in NightscoutUploadKit

Kdisimone:


ps2:
Thanks @Kdisimone  and @kenstack!

Thinking about moving this into treatments at some point, so it's not tied to device status updates, which are generally every 5 minutes.

PR #534, opened 2019-08-20
new Pod Settings buttons to control and test beeping

itsmojo:
[Enable/Disable Confirmation Beeps] [Enable/Disable Optional Pod Alarms] [Check Beeps]
Optional cancelNone() which will verifies the nonce to replace getStatus(),
flash logs read support, beep on Pod startup completion to verify piezo


OmniKit/MessageTransport/MessageBlocks/BasalScheduleExtraCommand.swift
```diff
@@ -62,7 +62,7 @@ public struct BasalScheduleExtraCommand : MessageBlock {
         rateEntries = entries
     }
     
-    public init(acknowledgementBeep: Bool, completionBeep: Bool, programReminderInterval: TimeInterval, currentEntryIndex: UInt8, remainingPulses: Double, delayUntilNextTenthOfPulse: TimeInterval, rateEntries: [RateEntry]) {
+    public init(acknowledgementBeep: Bool = false, completionBeep: Bool = false, programReminderInterval: TimeInterval = 0, currentEntryIndex: UInt8, remainingPulses: Double, delayUntilNextTenthOfPulse: TimeInterval, rateEntries: [RateEntry]) {
```
ps2: Parameters with defaults should go at the end of the function's parameter list.

OmniKit/MessageTransport/MessageBlocks/PodInfoFlashLogRecent.swift
```diff
@@ -8,24 +8,61 @@
 
 import Foundation
 
+let maxPumpEntriesReturned = 50
```
ps2: fileprivate

OmniKit/Model/AlertSlot.swift
```diff
@@ -99,13 +99,24 @@ public enum PodAlert: CustomStringConvertible, RawRepresentable, Equatable {
         case .finishSetupReminder:
             return AlertConfiguration(alertType: .slot7, duration: .minutes(55), trigger: .timeUntilAlert(.minutes(5)), beepRepeat: .every5Minutes, beepType: .bipBeepBipBeepBipBeepBipBeep)
         case .expirationAlert(let alertTime):
-            return AlertConfiguration(alertType: .slot3, duration: 0, trigger: .timeUntilAlert(alertTime), beepRepeat: .every1MinuteFor3MinutesAndRepeatEvery15Minutes, beepType: .bipBeepBipBeepBipBeepBipBeep)
+            if alertTime == 0 {
```
ps2:
```
let active = alertTime != 0 // disable if alertTime is 0
return AlertConfiguration(alertType: .slot3, active: active, duration: 0, trigger: .timeUntilAlert(alertTime), beepRepeat: .every1MinuteFor3MinutesAndRepeatEvery15Minutes, beepType: .bipBeepBipBeepBipBeepBipBeep)
```
This makes it clear how the check is being used and saves the reader looking through all the other arguments to see if there is anything else different.

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -31,6 +31,15 @@ public enum OmnipodPumpManagerError: Error {
     case notReadyForCannulaInsertion
 }
 
+public enum RunCommandSessionType {
```
ps2: I understand that you're trying to reduce the number of functions (and all the boiler plate that goes with them). But the savings comes at the expense of extensibility. If any of these commands later evolve to require parameters, or worse, want to return a different type, then it starts to get messy. I recommend keeping a function based api on the pump manager.

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -137,13 +137,21 @@ extension PodCommsError: LocalizedError {
     }
 }
 
-
+public var confirmationBeeps: Bool = false // TODO determine placement, create methods to get/set, make persistent?
```
ps2: These belong on OmnipodPumpManagerState

OmniKitUI/ViewControllers/OmnipodSettingsViewController.swift
```diff
@@ -282,6 +319,17 @@ class OmnipodSettingsViewController: RileyLinkSettingsViewController {
...
             case .suspendResume:
                 return suspendResumeTableViewCell
+            case .testCommand:
+                let cell = tableView.dequeueReusableCell(withIdentifier: SettingsTableViewCell.className, for: indexPath)
+                cell.textLabel?.text = LocalizedString("Test Command", comment: "The title of the command to run the test command")
+                cell.accessoryType = .disclosureIndicator
+                return cell
+            case .checkBeeps:
+                let cell = tableView.dequeueReusableCell(withIdentifier: SettingsTableViewCell.className, for: indexPath)
+                cell.textLabel?.text = LocalizedString("Check Beeps", comment: "The title of the command to run check beeps")
```
ps2: "Check Beeps" is somewhat ambiguous. My first assumption was that it reported the status of whether beeps were enabled. This should indicate it will play a sound on the pod.  Maybe "Emit Pod Beep", or "Play Test Beep"

ps2 (changes requested):
Thanks for submitting this early, Joe. Please see my comments on how we can get this merged. I'd be happy to take a look at the UI once the implementation detail comments have been addressed, if you wish.

OmniKit/MessageTransport/MessageBlocks/BasalScheduleExtraCommand.swift
itsmojo: Fixed

OmniKit/MessageTransport/MessageBlocks/PodInfoFlashLogRecent.swift
itsmojo: Done

OmniKit/Model/AlertSlot.swift
itsmojo: Done

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: Done, although I have some reservations about doing it this way.

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: Not implemented due to lack of time

OmniKitUI/ViewControllers/OmnipodSettingsViewController.swift
itsmojo: Check Beeps renamed to Play Test Beeps

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -862,7 +887,121 @@ extension OmnipodPumpManager {
...
+            case .failure(let error):
+                completion(error)
+            }
+        }
+    }
+
+    public func enableConfirmationBeeps(completion: @escaping (Error?) -> Void) {
+        self.confirmationBeeps = true
+        guard self.hasActivePod else {
+            completion(OmnipodPumpManagerError.noPodPaired)
```
ps2: Seems like this shouldn't be an error. It should be ok to enable beeps w/o a pod?

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -862,7 +887,121 @@ extension OmnipodPumpManager {
...
+                } catch let error {
+                    completion(error)
+                }
+            case .failure(let error):
+                completion(error)
+            }
+        }
+    }
+
+    public func disableConfirmationBeeps(completion: @escaping (Error?) -> Void) {
```
ps2: This function and the one above could be combined into a `setConfirmationBeeps(enabled: Bool)` function, since there are only two lines differing.

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -862,7 +887,121 @@ extension OmnipodPumpManager {
...
+                } catch let error {
+                    completion(error)
+                }
+            case .failure(let error):
+                completion(error)
+            }
+        }
+    }
+
+    public func disableOptionalPodAlarms(completion: @escaping (Error?) -> Void) {
```
ps2: Maybe combine with above function?

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -481,6 +502,12 @@ public class PodCommsSession {
 
             podState.updateFromStatusResponse(status)
 
+            if deliveryType == .all && basalBeeps && confirmationBeeps {
+                // With a deliveryType of .all and a beepType other than .noBeep, the Omnipod will emit 3 beeps!
```
ps2: This function shouldn't create a new way of beeping, issuing another command that could fail. A successful suspend with a failed beep emit will mark this suspend as uncertain, which is not ok.
Considering this with the fact that this alternate way of issuing beeps is making the api confusing, I think this logic should be moved to the PumpManager layer, if it's kept at all. The primary use case this is serving is bolus confirmation. And I'd be happy to take a PR that just addressed that; would be a lot simpler, for sure!

OmniKit/PumpManager/OmnipodPumpManagerState.swift
```diff
@@ -127,6 +133,18 @@ public struct OmnipodPumpManagerState: RawRepresentable, Equatable {
         } else {
             self.unstoredDoses = []
         }
+
+        if let rawConfirmationBeeps = rawValue["confirmationBeeps"] as? Bool {
+            self.confirmationBeeps = rawConfirmationBeeps
```
ps2: Maybe `self.confirmationBeeps = rawValue["confirmationBeeps"] as? Bool ?? false`?

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -540,27 +567,90 @@ public class PodCommsSession {
...
+    }
+
     public func getStatus() throws -> StatusResponse {
-        let response: StatusResponse = try send([GetStatusCommand()])
-        podState.updateFromStatusResponse(response)
-        return response
+        var statusResponse: StatusResponse
+
+        if useCancelNoneForStatus {
+            statusResponse = try cancelNone() // handles update from status response
```
ps2: Any reason this path isn't calling `updateFromStatusResponse`?

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -326,8 +324,31 @@ public class PodCommsSession {
         podState.updateFromStatusResponse(status)
         return status
     }
-    
-    public func insertCannula() throws -> TimeInterval {
+
+    // emits the specified beep type and sets the completion beep flags based on the specified confirmationBeep value
+    public func emitBeep(beepType: BeepType, confirmationBeeps: Bool) throws {
```
ps2: The function's name should be changed to communicate that it is reconfiguring a setting on the pod.

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -445,8 +466,8 @@ public class PodCommsSession {
         }
     }
     
-    public func cancelDelivery(deliveryType: CancelDeliveryCommand.DeliveryType, beepType: BeepType) -> CancelDeliveryResult {
-        
+    public func cancelDelivery(deliveryType: CancelDeliveryCommand.DeliveryType, beepType: BeepType, confirmationBeeps: Bool) -> CancelDeliveryResult {
```
ps2: This is a confusing api.  What does it mean to set beepType to `.none`, but confirmation beeps to `true`? Or with beepType set to some kind of beep, but confirmation beeps set to `false`?

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -606,6 +696,51 @@ public class PodCommsSession {
     public func assertOnSessionQueue() {
         transport.assertOnSessionQueue()
     }
+
+    public func enableConfirmationBeeps() throws {
```
ps2: Does these use cases warrant two new functions?

ps2 (changes requested):
Thanks Joe. Definitely moving in the right direction. Main issue I see is the cancelDelivery adding another call that can fail and change dosing certainty in an invalid way.  Some other smaller changes for clarity.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: {en,dis}ableConfirmationBeeps reworked to no longer report an error w/o a pod

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: Implemented in latest update

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: {enable,disable}OptionalPodAlarms are now combined into a single function with an argument

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: I think that we really should have a confirmation beep when suspending to match the one when resuming when Confirmation Beeps is enabled and basalBeeps is set. Unfortunately the Pod’s cancel delivery command can’t be used to do this as you’d get 3 sets of beeps in this case. I put this functionality into cancelDelivery() because it effectively implements what I think the Pod really should have done. For now decided to just remove this code and document as a TODO that suspending a Pod needs to have a single confirmation beep to match the one when resuming the Pod when Confirmation Beeps are enabled and basalBeeps have been configured

OmniKit/PumpManager/OmnipodPumpManagerState.swift
itsmojo: Done

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: Yes and that is what the comment was explaining. I have since restructured this routine to make this even more obvious.

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: emitBeep() has been renamed to beepConfig()

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: Agreed, but I blame a lot of this on the silly Pod cancelDelivery() semantics. beepType is the value to given to the Pod’s Cancel Delivery command and confirmationBeeps is needed both to indicate a confirmation beep is desired and to call beepConfig() to actually implement the confirmation beep for the deliveryType = .all case.

It the last this implementation (which has cancelDelivery providing a more logical cancel delivery for deliveryType: .all with confirmation beeps), when beepType is .none, confirmationBeeps is true, and deliveryType is .all, this would generate a single confirmation beep using the Pod beepConfig command. If beepType is some kind of beep, then you get whatever the Pod gives you and there will be no additional confirmation beep provided.

I have removed the confirmationBeeps argument to cancelDelivery() for now as I have decided to punt on optional resume delivery confirmation beeps for the time being to hopefully not delay things any further.

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: {enable,disable}ConfirmationBeeps are now combined to setConfirmationBeeps() which takes an argument

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -862,7 +893,91 @@ extension OmnipodPumpManager {
...
+                    completion(error)
+                }
+            case .failure(let error):
+                completion(error)
+            }
+        }
+    }
+
+    public func setConfirmationBeeps(enabled: Bool, completion: @escaping (Error?) -> Void) {
+        self.confirmationBeeps = enabled
```
ps2: If this command fails, it will report an error to the user, but will still have confirmation beeps enabled.

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -862,7 +893,91 @@ extension OmnipodPumpManager {
...
+                    completion(error)
+                }
+            case .failure(let error):
+                completion(error)
+            }
+        }
+    }
+
+    public func setOptionalPodAlarms(enabled: Bool, completion: @escaping (Error?) -> Void) {
+        self.optionalPodAlarms = enabled
```
ps2: Similarly to setConfirmationBeeps, we should probably only set this if we return a success condition.

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -326,8 +324,34 @@ public class PodCommsSession {
         podState.updateFromStatusResponse(status)
         return status
     }
-    
-    public func insertCannula() throws -> TimeInterval {
+
+    // emits the specified beep type and sets the completion beep flags based on the specified confirmationBeep value
+    public func beepConfig(beepConfigType: BeepConfigType, confirmationBeeps: Bool) throws {
```
ps2: This function should take `basalBeeps`, `tempBasalBeeps`, and `bolusBeeps` as arguments, instead of referring to global variables.

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -326,8 +324,34 @@ public class PodCommsSession {
...
+        let statusResponse: StatusResponse = try send([beepConfigCommand])
+        podState.updateFromStatusResponse(statusResponse)
+    }
+
+    private func configureCompleted(confirmationBeeps: Bool, optionalPodAlarms: Bool) throws {
+        podState.setupProgress = .completed
+
+        if optionalPodAlarms {
+            // program optional Pod alarms, will beep upon completion if supplementaryBeeps && confirmationBeeps
+            try enableOptionalPodAlarms(confirmationBeeps: confirmationBeeps)
```
ps2: If these fail, the exceptions will be thrown back to insertCannula, and reported as an error to the user, even though the setupProgress is .completed.  The UI will offer a retry to the user, which will then fail, as the pod is not in the right state.  If you want to consider this successful, I'd recommend swallowing the exceptions for the beeps here.

OmniKit/Model/BeepType.swift
```diff
@@ -19,10 +19,24 @@ public enum BeepType: UInt8 {
...
+    case bipBipBipbipBipBip = 0x7
+    case beeepBeeep = 0x8
+    // 0x9 and 0xA always return an error response for Beep Config
     case beepBeep = 0xB
     case beeep = 0xC
     case bipBeeeeep = 0xD
-    case fiveSecondBeep = 0xE    // 5 second alarm beep *if* Pod is currently suspended
-    case beepConfig_NoBeep = 0xF // For Beep Config Command no beep, else fatal Pod fault $37
+    case fiveSecondBeep = 0xE    // can only be used if Pod is currently suspended
+    case beepConfig_NoBeep = 0xF
```
ps2: `beepConfig_NoBeep` -> `noBeep`

ps2 (commented):
The global variables `basalBeeps`, `tempBasalBeeps` and `bolusBeeps` are being used a lot of different code paths, and conditionals. I'm not comfortable having such a wide reaching set of effects added to the code using global variables.  If these are features we want to support, then they need to be added in a maintainable manner.

ps2:
Work ongoing in separate PR

PR #545, opened 2019-09-09
Reworked confirmationBeep and optionalPodAlarms logic

itsmojo:
Reworked optional pod alarms to allow for eventual integration with user adjustable settings
Just have Play Test Beeps play the single type long beep sequence


itsmojo:
Abandon this branch

PR #548, opened 2019-09-11
generalize Bolus Beeps to Confirmation Beeps

itsmojo:
adjust Play Test Beeps to only emit one beep sequence
tweaks to better accomodate some actions on faulted pods
updated and improved logging
move readFlashLogs() to better integrate with confirmation beeps


itsmojo:
Abandon this branch

PR #553, opened 2019-09-27
Remote Overrides support

ps2:


ps2:
Also fixes https://github.com/LoopKit/Loop/issues/977

PR #554, opened 2019-09-28
Adjust activatedAt (expiresAt) on responses to account for clock drift

itsmojo:
Round alert configuration times to the nearest minute


ps2:
Thanks Joe!

PR #564, opened 2019-11-11
Detect remote flag for commands

ps2:
Adds decoding for remote request flag for suspend, resume, bolus, and temp basal events.


ecc1:
Does this work on x22 pumps? I may be missing something, but I think https://github.com/ps2/rileylink_ios/blob/remote-command-flag/MinimedKit/PumpEvents/BolusNormalPumpEvent.swift#L90 will reference an out-of-bounds byte in that case

ps2:
Thanks @ecc1, yes boluses were missing the proper handling for x22.

PR #565, opened 2019-11-13
Upgrade Bolus Beeps to more general purpose Confirmation Beeps

itsmojo:
which are Loop's version of the PDM's Confidence Reminders functionality
that when enabled adds confirmation beeps for user initiated Pod activity when doing
Suspend/Resume Delivery, Change Time Zone, Change Basal Rate, Test Command &
upon completion of Pod Setup.

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -13,6 +13,10 @@ import RileyLinkBLEKit
 import UserNotifications
 import os.log
 
+fileprivate let bolusConfirmationBeeps: Bool = true      // whether to emit bolus confirmation beeps
+fileprivate let basalConfirmationBeeps: Bool = true      // whether to emit basal confirmation beeps
+fileprivate let supplementaryBeeps: Bool = true          // whether to emit supplementary confirmation beeps
+fileprivate let tempBasalConfirmationBeeps: Bool = false // whether to emit temp basal confirmation beeps (for testing use)
 
 
 public enum ReservoirAlertState {
@@ -365,13 +369,13 @@ extension OmnipodPumpManager {
     }
 
     // Thread-safe
-    public var bolusBeeps: Bool {
+    public var confirmationBeeps: Bool {
         get {
-            return state.bolusBeeps
+            return state.confirmationBeeps
         }
         set {
             setState { (state) in
-                state.bolusBeeps = newValue
+                state.confirmationBeeps = newValue
             }
         }
     }
@@ -636,7 +640,16 @@ extension OmnipodPumpManager {
         #endif
     }
 
-    private func checkCannulaInsertionFinished() {
+    private func emitConfirmationBeep(session: PodCommsSession, beepConfigType: BeepConfigType) {
+        if self.confirmationBeeps && supplementaryBeeps {
+            let basalCompletionBeep = basalConfirmationBeeps
+            let tempBasalCompletionBeep = tempBasalConfirmationBeeps
+            let bolusCompletionBeep = bolusConfirmationBeeps
+            session.beepConfig(beepConfigType: beepConfigType, basalCompletionBeep: basalCompletionBeep, tempBasalCompletionBeep: tempBasalCompletionBeep, bolusCompletionBeep: bolusCompletionBeep)
+        }
+    }
+
+    public func checkCannulaInsertionFinished() {
         let deviceSelector = self.rileyLinkDeviceProvider.firstConnectedDevice
         self.podComms.runSession(withName: "Check cannula insertion finished", using: deviceSelector) { (result) in
             switch result {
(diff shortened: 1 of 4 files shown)


ps2:
I don't think I will be able to merge this; in general we want Loop to be something you use as little as possible, and when you do use it, you shouldn't be wondering if it's doing what you told it, and then waiting to confirm that something happened. Changes that help bring Loop to the point where users can rely on Loop to tell them if something didn't happen would be more welcome. Those kinds of changes give the user more confidence in Loop and let them set down Loop faster.

itsmojo:
Regarding “we want Loop to be something you use as little as possible, and when you do use it, you shouldn't be wondering if it's doing what you told it, and then waiting to confirm that something happened”, unless I'm misunderstanding this appears to actually be a great reason why Confirmation Beeps should be an available option. Without confirmation beeps, a user doing something like a Suspend Delivery, Resume Delivery, Change Time Zone or Sync’ing of the Basal Rates *does* has to wait to watching the Loop app to confirm that the requested action happened. While with Confirmation Beeps a user *doesn’t* have to wait and can look away to just get a gentle audio beep to confirm that something happened. Confirmation beeps is exactly a kind of change that give the user more confidence in Loop and lets them set down Loop faster and so it should be a welcome change! QED

Confirmation beeps are Loop’s equivalent to the PDM’s Confidence Reminders which many PDM users like to use. It’s not really that much more than the existing Bolus Beeps which is giving optional audio confirmation for the most common manual Pod interaction,  giving boluses. Why exclude other basic operations of suspending, resuming, setting a basal schedule or changing the time zone from also having an optional audio confirmation as well?

I have been told by a variety of users that having confirmation beeps would be very helpful for beginners, younger children, more elderly adults, and part time care givers & school nurses who are naturally more nervous or unsure about DIY Loop and who can always use and appreciate items that can give them more confidence when manually enacting something on Loop. With a traditional tubed insulin pump there is visual confirmation feedback available on the pump screen itself, but with a Pod the only real feedback mechanism available on the device itself is using confirmation beeps.

Additionally, having confirmation beeps can be critical giving users feedback that a Pod operation was manually enacted when it wasn’t actually intended. I have heard from elderly Loop users who repeatedly accidentally suspend their pump without realizing it when they had unknowingly pressed the “Suspend Delivery” button while scrolling. Many times they don’t know that their pump had been suspended until much later when they notice that there a red loop and then they finally figure out that they must have accidentally suspended their pump.

Kdisimone:
> ...you shouldn't be wondering if it's doing what you told it, and then waiting to confirm that something happened...

Sorry...premature return

Kdisimone:
Agreed...but I'm seeing a need for these confirmation beeps in my private messages and generally in facebook posts.  The use case for confirmation beeps for basal changes and suspend/resume is big with school nurses.  They (school nurses) are more nervous by nature with Loop than parents are...they feel the pressure of professional liability in caring for kids, feel like DIY is already a stretch in comfort, don't want to be "blamed" if things go wrong, and generally will not have a chance to develop a "trust" in a system simply because their interactions will be much fewer and farther between than a parent's experience is.   

There's a lot of parents fighting school nurses already with a tenuous agreement to loop, and I think confirmation beeps would go a very long way to helping smooth out some of that lack of confidence that keeps nurses from feeling comfortable in some situations. A quick beep will let them "know" with a more weight of evidence and a stronger sense of confidence that their action has been implemented. Bolus beeps have been very very well received by that user group and confirmation beeps for basal changes and suspend/resume would be welcome for the same reasons.  

I'd really hope we can get these confirmation beeps in. I'd be happy to ask a couple parents who have reached out to me to chime in here. I've set some up already on this branch just because of their school nurse experiences being improved through confirmation beeps and they were hoping to get more than boluses.

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -13,6 +13,10 @@ import RileyLinkBLEKit
 import UserNotifications
 import os.log
 
+fileprivate let bolusConfirmationBeeps: Bool = true      // whether to emit bolus confirmation beeps
```
ps2: Please remove these vars. If we want to make various options available to users, they should be as UI elements; otherwise, lets pick a set that works for most people and use that.  The latter is what I'd prefer.  Based on feedback, it seems like Bolus, and suspend/resume, and basal change beeps fit that mark.

ps2 (changes requested):

ps2:
Thanks for sharing Katie.  I have a couple comments, and then I think we can get this merged.

itsmojo:
The last commit removes the fileprivate variables that allow customization of the confirmation beeping for bolus, basal and supplementary operations which I really wanted to leave in place as this was never supposed to be a standard user operation, just something available for developers or hardcore types who might want to customize their confirmation beeps experience. I left the variable for temp basals as this is an important testing/debug option to have available to manually enable for developers and is not obvious for a developer who might want to enable confirmation beeps for TBs. Optional confirmation beeps are definitely needed for the very long read pulse log operation. Having them also available for the manual read pod status command will make every manual operation involving pod communications have an optional confirmation beep which provides consistency and increased user confidence in Loop’s control of the Pod.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: All these variables are now removed but one. See additional comments on recent commit for details.

PR #567, opened 2019-11-22
Calculate insertion wait time from Pod cannulaInsertionUnits constant

itsmojo:
Instead of have a fixed 10 second value for insertionWait constant,
calculate this value as (Pod.cannulaInsertionUnits/Pod.pulseSize)
(i.e., 0.5/0.05 = 10 with default cannulaInsertionUnits) so
that the insertion wait time will scale with cannulaInsertionUnits

### DIFF
--- a/OmniKit/MessageTransport/MessageBlocks/BolusExtraCommand.swift
+++ b/OmniKit/MessageTransport/MessageBlocks/BolusExtraCommand.swift
@@ -64,7 +64,7 @@ public struct BolusExtraCommand : MessageBlock {
         squareWaveDuration = timeBetweenExtendedPulses * Double(pulseCountX10) / 10
     }
     
-    public init(units: Double, timeBetweenPulses: TimeInterval = 2, squareWaveUnits: Double = 0.0, squareWaveDuration: TimeInterval = 0, acknowledgementBeep: Bool = false, completionBeep: Bool = false, programReminderInterval: TimeInterval = 0) {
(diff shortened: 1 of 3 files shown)


marionbarker:
I request that this be merged when convenient.  Per Joe's suggestion, I customized my cannulaInsertionUnits value in Pod.swift.  Without this change, Loop thinks insertion did not complete (because it took more than 10 sec) and I had to go through 3 additional screens following a successful insertion screen: "complete pod setup", insert cannula and then an immediate insertion successful screen.  (names approximate). I've updated my personal version to have this change and it fixed the problem. Fortunately, pods all worked fine but it was a little disconcerting - especially the first time.

ps2:
`Pod.pulseSize` is not the right divisor; it should be a rate. It works here because it just happens to be the same as cannula insertion units per second.  I.e. if the pod pulse size was a 0.025 U size, the math would be wrong. We should have a variable on `Pod` that represents the cannula dosing rate. Maybe `primeUnitsPerSecond` to share it with prime, since they share the same rate, or make a separate one `cannulaInsertionUnitsPerSecond`.

itsmojo:
The math is actually fully correct independent of the pod pulse size, but having a constant for a rate using the prime pulse timing makes things more understandable & consistent. Created a new Pod constant for primeDeliveryRate (to match the existing bolusDeliveryRate constant) as well for secondsPerPrimePulse and secondsPerBolusPulse and reworked derived values and code to use these new constants as appropriate.

PR #570, opened 2019-12-02
Remove CA region from selection

achkars:
When selecting pump regions, NA and CA both return North America on UI.
Causes confusion while selecting.
Changed NA to NA/CA in picker.


ps2:
I believe this is to help users who have a pump that the region code on the back of the pump is "CA".  For both pump region codes "CA" and "NA", we use the "North America" tuning.

achkars:
That's correct.

ps2:
I think the CA/NA distinction should be kept to let users of both pump types know what to input, even if it does end up with the same tuning.

achkars:
Created a new PR #573 with separate CA and NA regions.

PR #577, opened 2019-12-29
Add new Read Pulse Log command to Pod Settings

itsmojo:
* Display the read pulse log entries with headers
* Change all flash log labels to pulse log instead

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -1185,18 +1185,14 @@ extension OmnipodPumpManager: PumpManager {
                 state.suspendEngageState = .engaging
             })
 
-            // N.B. with a deliveryType of .all and a beepType other then .noBeep, the Pod will emit 3 beeps! Use .noBeep here & do beeping at end.
-            let result = session.cancelDelivery(deliveryType: .all, beepType: .noBeep)
+            let beepType: BeepType = self.confirmationBeeps ? .beeeeeep : .noBeep
+            let result = session.cancelDelivery(deliveryType: .all, beepType: beepType)
             switch result {
             case .certainFailure(let error):
                 completion(error)
             case .uncertainFailure(let error):
                 completion(error)
             case .success:
-                // Do a separate single confirmation beep if appropriate. There are no in-progress deliveries to worry about after the cancel all.
-                if self.confirmationBeeps {
-                    session.beepConfig(beepConfigType: .beeeeeep, basalCompletionBeep: false, tempBasalCompletionBeep: false, bolusCompletionBeep: false)
-                }
                 session.dosesForStorage() { (doses) -> Bool in
                     return self.store(doses: doses, in: session)
                 }
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
-github "LoopKit/LoopKit" "dev"
+github "LoopKit/LoopKit" ~> 3.0
 github "LoopKit/MKRingProgressView" "appex-safe"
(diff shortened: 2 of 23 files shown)


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -81,6 +81,23 @@ extension OmnipodPumpManagerError: LocalizedError {
     }
 }
 
+extension BinaryInteger {
```
ps2: This extension should be fileprivate in the same file responsible for decoding the pulse log (PodInfoPulseLogRecent.swift).

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1006,38 +1023,56 @@ extension OmnipodPumpManager {
         }
     }
 
-    public func readFlashLog(completion: @escaping (Error?) -> Void) {
-        // use hasSetupPod to be able to read the flash log from a faulted Pod
+    private func pulseLogString(pulseLogEntries: [UInt32], lastPulseNumber: Int) -> String {
```
ps2: This should be in PodInfoPulseLogRecent, since it's a serialization of that object.

OmniKitTests/PodInfoTests.swift
```diff
@@ -343,27 +343,125 @@ class PodInfoTests: XCTestCase {
...
+            let decoded = try PodInfoPulseLogPrevious(encodedData: Data(hexadecimalString: "51003214602500196128001c6124002161280024612500296129002c60260031602a003460260039612a803c61268041602c800060278005632880086025800d6128801061258015612780186023801d6026802061228025602700286124002d2128003020270035202a00382027003d202a004020290045202c0048202a004d212c005021290055212c00582129805d202b806020288001202d8004212a8009202d800c21298011212a80142129801921801c212a8021212c8024202c0029212f002c212d003120310082")!)
+            XCTAssertEqual(.dumpOlderPulseLog, decoded.podInfoType)
             XCTAssertEqual(50, decoded.nEntries)
-            XCTAssertEqual(Data(hexadecimalString:"14602500196128001c6124002161280024612500296129002c60260031602a003460260039612a803c61268041602c800060278005632880086025800d6128801061258015612780186023801d6026802061228025602700286124002d2128003020270035202a00382027003d202a004020290045202c0048202a004d212c005021290055212c00582129805d202b806020288001202d8004212a8009202d800c21298011212a80142129801921801c212a8021212c8024202c0029212f002c212d003120310082"), decoded.hexWordLog)
+            XCTAssertEqual(0x14602500, decoded.pulseLogEntry[0])
```
ps2: If the same code is decoding entry[0] as entry[n], it's probably sufficient to just test any variations that exercise different aspects of the decoding.

OmniKit/MessageTransport/MessageBlocks/PodInfoFlashLogRecent.swift
```diff
@@ -10,59 +10,67 @@ import Foundation
...
+            index += 1
+        }
     }
-    // TODO add code to nicely format the 32-bit pump log entries
 }
 
-// read (up to) the most previous 50 32-bit pump entries from flash log
-public struct PodInfoFlashLogPrevious : PodInfo {
+// read (up to) the most previous 50 32-bit pump entries from pulse log
+public struct PodInfoPulseLogPrevious : PodInfo {
```
ps2: This seems to be largely a copy of the previous decoding. Can they be represented by the same decoding and the same structure?

ps2 (changes requested):
Just some organization recommendations and question about some copied code.

ps2:
What is different between PodInfoPulseLogRecent and PodInfoPulseLogPrevious? Seems like they could be combined into a single struct.

itsmojo:
The PodInfoPulseLogRecent has information about what the pulse # is for pulse log entries (bytes 3:4 called IIII) while PodInfoPulseLogPrevious has no such information (bytes 3:4 called NNNN is really just redundant length info). This means that to understand what the pulse #'s are for the pulse log entries in a PodInfoPulseLogPrevious return, you need to have the pulse # context which is contained only in the PodInfoPulseLogRecent return. It seems silly that Insulet defined the PodInfoPulseLogPrevious (pod info type 0x51) this way.

ps2 (approved):

ps2:
Already merged; not sure why the PR is still open. Closing.

PR #581, opened 2020-01-16
Rework cancelDelivery to provide reasonable beep semantics

itsmojo:
* Use two cancel commands in a single message to implement an integrated single
suspend confirmation beep instead of having to use a separate beepConfig message
* Various confirmation beep related cleanup items


ps2 (approved):
Just one question: by splitting up the cancel into two commands, is there any chance that one of them will succeed and the other will fail?

itsmojo:
For a two cancel commands in a single message, the only way that the first one would success and the second one would not that I can imagine from looking at the firmware would be if two different nonces were used and the second nonce was invalid OR if an invalid beep nibble value was used for the 2nd cancel command. Since we use the same nonce for both of the cancel commands (and only a valid beep nibble can be specified by the Swift code), then both cancel commands would either succeed or fail based on the validity of the nonce value used.

PR #588, opened 2020-03-17
Fix for pod pairing issue with accepting wrong response

ps2:
* Only accept responses with the sequence id following that of the sent message.
* Rename commands and functions from "configure pod" to "setup pod", to unify, and match openomni wiki.
* Add error message description for unexpected ack response
* Properly increment packet and sequence numbers over first 3 message exchanges during pairing.


ps2:
Thanks to @itsmojo for identifying the issue with Loop accepting the wrong packet.

PR #589, opened 2020-03-22
Use submessages for confirmation beeps instead of a separate message

itsmojo:
Fix Play Test Beeps so that it can report failures
Report errors for confirmation beeps on Read Pulse Log
Don't change the confirmationBeeps variable on beepConfig error
Use new improved beep submessage cancel method when suspending
Add confirmation beeping back for Read Pod Status
Add confirmation beeping for acknowledge alerts
send() now has optional prefixMessage and appendMessage submessages
Improved comments for possible send() errors
Remove obsolete comment in prime() for old PDM action
Improved comments for prime() actions and pod state
beepConfig() now returns optional StatusResponse and can throw
Remove previous split cancel command workaround to handle suspend beeping

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -638,10 +638,15 @@ extension OmnipodPumpManager {
         #endif
     }
 
-    private func emitConfirmationBeep(session: PodCommsSession, beepConfigType: BeepConfigType) {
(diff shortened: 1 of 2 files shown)


itsmojo:
Open issue for some of user visible problems that will be fixed with this PR: https://github.com/ps2/rileylink_ios/issues/616

ps2:
@itsmojo That link doesn't seem to work?

PR #590, opened 2020-03-24
Optionally pad messages containing a nonce to span a packet

itsmojo:


OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -175,6 +175,21 @@ public class PodCommsSession {
...
     }
 
+    // pad the given message to be more than one packet using GetStatus sub-messages
+    private func padMessage(message: [MessageBlock]) -> [MessageBlock] {
+        let packetLen = PacketType.pdm.maxBodyLen()
+        let getStatusCommand = GetStatusCommand()
+        let getStatusLen = 3
+
+        var paddedMessage = message
+        var encodedLen = messageLength(message: message)
```
ps2:
```

File: OmniKit/MessageTransport/Message.swift [hunk 2]
```diff
@@ -100,6 +100,13 @@ struct Message {
     }
 }
 
+// returns the encoded length of a message
+public func messageLength(message: [MessageBlock]) -> Int {
```
ps2: This should be a static func on the Message class, or just replace it with an inline version as suggested below.

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -175,6 +175,21 @@ public class PodCommsSession {
...
+        while encodedLen <= packetLen {
```
ps2, line 186: Are there commands short enough to need more than one status added to them to reach two packets?

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -194,6 +209,10 @@ public class PodCommsSession {
         
         if blocksToSend.contains(where: { $0 as? NonceResyncableMessageBlock != nil }) {
             podState.advanceToNextNonce()
+            let padNonceMessages = false // edit to pad non deactivatePod nonce messages
+            if padNonceMessages && messageBlocks[0].blockType != .deactivatePod {
```
ps2, line 213: I don't think debug/dev code should linger in the code, but I think we should talk through whether we want to enable this for Loop in general.  Which nonce commands end up getting padded?

ps2 (commented):
Small suggestions, and a question.

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: Done

OmniKit/MessageTransport/Message.swift
itsmojo: Made static func on the Message class as I didn't see an inline version suggestion

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: The following nonce commands all need multiple 3-byte get status sub-commands to pad to reach two packets: 
$08 command is only 7 bytes long, the $11 command is 6 bytes long & the $1F command is 6 bytes long. The $19 command is a minimum of 12 bytes long, but its length is variable. The $1C deactivate pod command is only 5 bytes, but this command can't be padded (as is the case for all the $1A commands which never fit in a simple packet with the required $13, $16 or $17 sub-command). The total encoded MessageBlock length has to exceed 25 bytes (= 31 - 4 [address] - 1 [B9] - 1 [Blen]) to not fit in a single packet.

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: The $08, $11, $1F and most $19 commands would end up getting padded when this code is enabled. I'm happy to talk about whether the nonce command padding should be enabled for Loop in general any time! My vote at this time would be not to bother, but to have this code available in case anyone wants to do some testing with and without nonce command padding enabled.

ps2: Hmm, the only one that I could see being a significant benefit is the cancel delivery ($1F) command. But yeah, out of all the commands, bolus and temp basal are by far the most important to have this feature on, and those already have the feature of two packets. I am fine leaving this branch out there for future reference, and I think that we do have some other options in terms of recovering from an uncertain delivery state. Some of the code in this latest merge will support automatic recovery once the PumpManager side of it is implemented in OmniKit.

ps2:
I'm going to close this for now; the branch and code will still be there if we want to reopen, or reference it.

PR #595, opened 2020-04-10
Updated VersionResponse definition and added additional verifications

itsmojo:
Updated VersionResponse field names, ordering, and offsets to batter match wiki
VersionResponse address is never optional
VersionResponse has new optional gain and rssi for the shorter response
Added new isAssignAddressVersionResponse & isAssignAddressVersionResponse vars
Added additional verifications of returned VersionResponse during pairing
Added log.default message for pod info including signal strength during pairing
Added %{public} for all log.error sub-strings in PodComms
Updated commenting in both VersionResponse and PodComms

### DIFF
--- a/OmniKit/PumpManager/PodComms.swift
+++ b/OmniKit/PumpManager/PodComms.swift
@@ -49,7 +49,7 @@ class PodComms: CustomDebugStringConvertible {
         
         let messageTransportState = MessageTransportState(packetNumber: 0, messageNumber: 0)
         
-        // Create random address with 20 bits.  Can we use all 24 bits?
+        // Create random address with 20 bits to match PDM, could easily use 24 bits instead
         let address = 0x1f000000 | (arc4random() & 0x000fffff)
         
         let transport = PodMessageTransport(session: commandSession, address: 0xffffffff, ackAddress: address, state: messageTransportState)
@@ -63,14 +63,19 @@ class PodComms: CustomDebugStringConvertible {
         let response = try transport.sendMessage(message)
 
         if let fault = response.fault {
-            self.log.error("Pod Fault: %@", String(describing: fault))
+            self.log.error("Pod Fault: %{public}@", String(describing: fault))
             throw PodCommsError.podFault(fault: fault)
         }
         
-        guard let config = response.messageBlocks[0] as? VersionResponse else {
+        guard let config = response.messageBlocks[0] as? VersionResponse,
+            config.isAssignAddressVersionResponse == true
+        else {
+            self.log.error("assignAddress unexpected response: %{public}@", String(describing: response))
             let responseType = response.messageBlocks[0].blockType
             throw PodCommsError.unexpectedResponse(response: responseType)
         }
+
+        self.log.default("Assigned address 0x%x to pod lot %u tid %u, signal strength %u", config.address, config.lot, config.tid, config.rssi!)
         
         // Pairing state should be addressAssigned
         return PodState(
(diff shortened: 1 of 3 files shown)


OmniKit/PumpManager/PodComms.swift
```diff
@@ -112,11 +117,14 @@ class PodComms: CustomDebugStringConvertible {
...
 
         if let fault = response.fault {
-            self.log.error("Pod Fault: %@", String(describing: fault))
+            self.log.error("Pod Fault: %{public}@", String(describing: fault))
             throw PodCommsError.podFault(fault: fault)
         }
 
-        guard let config = response.messageBlocks[0] as? VersionResponse else {
+        guard let config = response.messageBlocks[0] as? VersionResponse,
+            config.isSetupPodVersionResponse == true
```
ps2, line 129: Pairing logic change here. Have we seen the pod return the shorter version here, and if so, what problems does it cause?

ps2: Wondering if we should just accept this if it happens; we will have all the info required to proceed.

itsmojo: We have seen a mismatched 01 version response when we had some transport errors that have since been fixed. The only correct response for an 03 SetupPod command is an 011b which requires use of both checks just as the only correct response for an 07 AssignAddress command is an 0115 which also requires both checks. In the pod firmware these are actually totally different code paths even though they both build similar looking, but actually different, 01 responses.

We should never accept the incorrect type of 01 VersionResponse or the pairing will mysteriously hang doing the next set up in pairing (as seen in last pairing failure that Eelke documented). Also, if the address/lot/tid don't match with the PodState info already saved, the pairing would mysteriously fail in the next step with either a no response or a bad nonce error.

itsmojo: If we get back an 011b response from the 03 SetupPod command which doesn't match the address/lot/tid previously saved from the (potentially several minutes) earlier 07 AssignAddress command, then we do have all the info required (and the pod could potentially be in a state) to continue if Loop recreated all its PodState to match. I went with a minimalistic approach and just detected and noted this condition as the pairing would mysteriously fail if we continued. Depending on the exact details, the user could possibly recover by simply moving the pods around to get the original pod back in range and retrying.

itsmojo: We have seen a mismatched 01 version response when we had some transport errors that have since been fixed. The only correct response for an 03 SetupPod command is an 011b which requires use of both checks just as the only correct response for an 07 AssignAddress command is an 0115 which also requires both checks. In the pod firmware these are actually totally different code paths even though they both build similar looking, but actually different, 01 responses.

We should never accept the incorrect type of 01 VersionResponse or the pairing will mysteriously hang doing the next set up in pairing (as seen in last pairing failure that Eelke documented). Also, if the address/lot/tid don't match with the PodState info already saved, the pairing would mysteriously fail in the next step with either a no response or a bad nonce error.

ps2: I think the address check, and lot/tid checks below are sufficient (and definitely needed at this stage, thanks!). This check seems to be an extraneous check. Please remove it and this PR is good to go. If you want to log it, that is fine, but it sounds like we don't expect this ever to happen, at least not for messages with the address we just assigned.

itsmojo: Verifying that we received the 011b response from the SetupPod Command is not an extraneous check. This is exactly what happened in Eelke’s latest pairing failure when there was a transport problem (i.e., message seq #’s weren’t being verified) and it lead to yet another pairing hang. Even though the 0115 and 011b responses use the same 01 response # and have many of the same fields (albeit at different offsets), the 0115 and 011b responses are very particular to the 07 assignAddress and 03 SetupPod commands respectively. It is true that (other than something like a pod fault) we should never ever get the wrong response for any command, but if we have another transport bug or some other problem where the wrong response is somehow returned, it’s best and proper to catch it as early we can. It’s just that for the 01 version responses, we need to treat an 0115 response as completely different than an 011b response. Getting anything other than an 011b response from a 03 SetupPod command is never correct as it would indicate that the Pod is not actually paired.

Separately I noticed that the setupState enum that was created just for VersionResponse should really be a PodProgressStatus. Earlier today I also noticed that some of the names of the Pod Progress values are confusing/incorrect. I have already updated the wiki, the firmware sources, and my private Loop code to use names and descriptions that are more consistent/correct. But to update the names in Loop would entail new strings to internationalize which would delay 2.1, so I decided not to create another PR (or updating this PR) to replace setupState in VersionResponse with a podProgressStatus and updating the names and strings in PodProgressStatus at this time.

PR #598, opened 2020-04-14
Accept packets with a packet address of 0xFFFFFFFF to match PDM & Pod

itsmojo:
To be used with PR #597: Verify the message address before accepting


ps2 (approved):
LGTM!

PR #600, opened 2020-04-29
Add additional checking before attempting a cancel command

itsmojo:
Don't attempt a cancel command before the pod setup has been complete to avoid
possible pod fault (pod progress >= 3 and <= 7) or no response (pod progress <= 2)

### DIFF
--- a/OmniKit/PumpManager/PodCommsSession.swift
+++ b/OmniKit/PumpManager/PodCommsSession.swift
@@ -457,11 +457,6 @@ public class PodCommsSession {
     public func cancelDelivery(deliveryType: CancelDeliveryCommand.DeliveryType, beepType: BeepType) -> CancelDeliveryResult {
         var message: [MessageBlock]
 
-        // Pod will fault or not respond if its setup is not complete, just return a noResponse error if command can't be run.
-        guard podState.setupProgress == .completed else {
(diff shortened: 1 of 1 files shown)


OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -457,6 +457,11 @@ public class PodCommsSession {
     public func cancelDelivery(deliveryType: CancelDeliveryCommand.DeliveryType, beepType: BeepType) -> CancelDeliveryResult {
         var message: [MessageBlock]
 
+        // Pod will fault or not respond if its setup is not complete, just return a noResponse error if command can't be run.
+        guard podState.setupProgress == .completed else {
+            return CancelDeliveryResult.certainFailure(error: PodCommsError.noResponse)
```
ps2: Returning a noResponse here for something we didn't attempt could create some confusion. Would rather you just remove this check, since the more sensical code path is to just not attempt in this state, which you've done below.

ps2 (changes requested):

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: Removed extra check in cancelDelivery itself to prevent a potential pod fault if cancelDelivery() is called incorrectly as requested

PR #601, opened 2020-04-30
Various error message fixes and improvements

itsmojo:
* Use the correct invalid address argument for PodCommsError.invalidAddress error
* Remove unneeded periods in new PodCommsError recoverySuggestion strings to avoid double periods
* Have errorDescription and recoverySuggestion switches in same order as the PodCommsError enum
* Verify that the assignAddress version response is actually from an assignAddress command

### DIFF
--- a/OmniKit/PumpManager/PodComms.swift
+++ b/OmniKit/PumpManager/PodComms.swift
@@ -67,8 +67,9 @@ class PodComms: CustomDebugStringConvertible {
             throw PodCommsError.podFault(fault: fault)
         }
         
-        guard let config = response.messageBlocks[0] as? VersionResponse else
-        {
+        guard let config = response.messageBlocks[0] as? VersionResponse,
+            config.isAssignAddressVersionResponse == true
+        else {
             self.log.error("assignAddress unexpected response: %{public}@", String(describing: response))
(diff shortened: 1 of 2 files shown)


OmniKit/PumpManager/PodComms.swift
```diff
@@ -67,16 +67,17 @@ class PodComms: CustomDebugStringConvertible {
             throw PodCommsError.podFault(fault: fault)
         }
         
-        guard let config = response.messageBlocks[0] as? VersionResponse else
-        {
+        guard let config = response.messageBlocks[0] as? VersionResponse,
+            config.isAssignAddressVersionResponse == true
```
ps2: I think this is overly defensive and not strictly necessary.

ps2 (changes requested):
Good fix for the config.address error message, and cleanup on the recovery messages. Just one small request to remove an unnecessary check.

OmniKit/PumpManager/PodComms.swift
itsmojo: I'm guarding against potential transport errors &/or pods in funky states which we've seen before. We should catch any none $0115 AssignAddress responses here instead of proceeding.

ps2 (approved):

Issue #618, opened 2020-09-14
Omnipod commands errors should display localized strings when possible


itsmojo:
Every Omnipod command in Pod Settings (except for the "Read Pod Status" command) will print more cryptic error messages instead of the more user friendly localized error description when available. This list includes the Suspend/Resume Delivery, Read Pulse Log, Test Command, Change Time Zone, Enable/Disable Confirmation Beeps commands. The "Play Test Beeps" command also has this same behavior, but currently it incorrectly always reports "Succeeded" for errors as reported in https://github.com/ps2/rileylink_ios/issues/616 ("Play Test Beeps errors reported as success plus various confirmation beeps issues") which will be fixed when https://github.com/ps2/rileylink_ios/pull/589 is eventually merged.

![Suspend-error-glop](https://user-images.githubusercontent.com/23020111/93042249-6ea71580-f603-11ea-914a-0bdd253d15fd.PNG)
![Resume-error-glop](https://user-images.githubusercontent.com/23020111/93042275-78307d80-f603-11ea-9543-365e2097d624.PNG)
![Read-Pulse-Log-error-glob](https://user-images.githubusercontent.com/23020111/93042295-84b4d600-f603-11ea-8af0-d0ba260495fc.PNG)
![Test-Command-error-glop](https://user-images.githubusercontent.com/23020111/93042312-8c747a80-f603-11ea-90c8-c10aeb63ff33.PNG)
![Change-Time-Zone-error-glop](https://user-images.githubusercontent.com/23020111/93042326-95fde280-f603-11ea-8488-1b382bfba0e2.PNG)
![Enabling-Confirmation-Beeps-error-glop](https://user-images.githubusercontent.com/23020111/93042581-479d1380-f604-11ea-81e0-95fc77fb9b65.PNG)
![Disabling-Confirmation-Beeps-error-glop](https://user-images.githubusercontent.com/23020111/93042589-4cfa5e00-f604-11ea-8811-15d3f4741e0a.PNG)

![Play-Test-Beep-error-glop-post-PR589](https://user-images.githubusercontent.com/23020111/93042859-0fe29b80-f605-11ea-9cb7-658391f336eb.PNG)

[...]

PR #620, opened 2020-09-18
Fix issue #618: Omnipod errors should display localized Strings

itsmojo:
* Return the underlying localized error string for PodCommError.commsError
* Have all OmnipodPumpManager commands return a localized error
* Fix coding errors in CommandResponseViewController for Read Pulse Log
and Read Pulse Log which return a String and not an error

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -947,9 +947,10 @@ extension OmnipodPumpManager {
         return result
     }
     
-    public func readPodStatus(completion: @escaping (String?) -> Void) {
-        guard self.hasActivePod else {
-            completion(String(describing: OmnipodPumpManagerError.noPodPaired))
+    public func readPodStatus(completion: @escaping (String) -> Void) {
+        // use hasSetupPod to be able to read pod info from a faulted Pod
+        guard self.hasSetupPod else {
+            completion(PodCommsError.noPodPaired.errorDescription!)
             return
         }
 
@@ -970,7 +971,7 @@ extension OmnipodPumpManager {
                     })
                     completion(self.podStatusString(status: status))
                 case .failure(let error):
-                    throw error
+                    reportError(String(describing: error))
                 }
             } catch let error as LocalizedError {
                 reportError(error.localizedDescription)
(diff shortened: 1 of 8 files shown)


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -947,9 +947,10 @@ extension OmnipodPumpManager {
         return result
     }
     
-    public func readPodStatus(completion: @escaping (String?) -> Void) {
-        guard self.hasActivePod else {
-            completion(String(describing: OmnipodPumpManagerError.noPodPaired))
+    public func readPodStatus(completion: @escaping (String) -> Void) {
```
ps2, line 950: It would be nice to change this to return Result<String, Error>, but we can do that later.

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1032,15 +1033,26 @@ extension OmnipodPumpManager {
         }
     }
 
-    public func readPulseLog(completion: @escaping (String?) -> Void) {
+    public func readPulseLog(completion: @escaping (String) -> Void) {
```
ps2, line 1034: Same here: Result<String, Error> would be more idiomatic.

OmniKitUI/ViewControllers/CommandResponseViewController.swift
```diff
@@ -82,7 +84,9 @@ extension CommandResponseViewController {
         return T { (completionHandler) -> String in
             pumpManager.playTestBeeps() { (error) in
                 let response: String
-                if let error = error {
+                if let localizedError = error as? LocalizedError, !localizedError.localizedDescription.isEmpty {
```
ps2: Just calling error.localizedDescription will work for this. Those PodCommErrors that return nil in errorDescription will return an odd string for `localizedDescription`, so the "isEmpty" will always be false. 

We can add descriptions for those PodCommErrors that have nil errorDescriptions later.

ps2 (changes requested):
Looks good in general. I think we should avoid the checking and downcasting of errors in general, and use localizedDescription for these types of errors that we're simply printing in the testing commands.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: My initial impression was that having a <String, Error> return for OmnipodPumpManager's ReadPodStatus and ReadPulseLog commands might make things more complicated than just returning a String to display (with either the results or the error message), but I would certainly be willing to rework these later if needed now that I better understand how this might look.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: Ditto

OmniKitUI/ViewControllers/CommandResponseViewController.swift
itsmojo: Thanks for the info. Updated code assuming that localizedError.localizedDescription.isEmpty will always be false

OmniKitUI/ViewControllers/CommandResponseViewController.swift
```diff
@@ -27,15 +27,11 @@ extension CommandResponseViewController {
...
-                            response = String(describing: error)
-                        } else {
-                            response = error.localizedDescription
-                        }
+                        response = error.localizedDescription
                     } else {
                         response = messageWithRecovery
                     }
-                } else if let localizedError = error as? LocalizedError, !localizedError.localizedDescription.isEmpty {
+                } else if let localizedError = error as? LocalizedError {
```
ps2: I think the whole conditional can be simplified:
```
} else if let localizedError = error as? LocalizedError {
    response = localizedError.localizedDescription
} else if let error = error {
    response = String(describing: error)
}
```
Can be turned into:
```
} else if let error = error {
    response = error.localizedDescription
}
```

itsmojo: Makes sense if localizedDescription.isEmpty is always false

ps2: I should've mentioned there are several of these in this PR that could be simplified.

itsmojo: Done. Now that PodCommsError.commsError.localizedDescription has a more meaningful content, the simplified version will display a meaningful error message for RileyLink errors with this more simplified method.

Issue #617, opened 2020-09-14
OmniKitUI displays a mystery single dot for certain pairing, cannula insertion, and replace pod errors


itsmojo:
In the following Zulip post, Alecia Yurkovich reported that she was getting a single "." when trying to pair using a RileyLink that had recently fallen into a pool. Upgrading from Loop 2.0 to 2.2 didn't fix the problem.

https://loop.zulipchat.com/#narrow/stream/144111-general/topic/Omnipod.20Pairing.20Issues/near/207637727

Loop displaying a single "." on a pairing error is the result of PairPodSetupViewController not generating an error string to display for the particular error being returned and then appending a "." to this empty string. PairPodSetupViewController should always display some appropriate & non-empty error message for any errors returned and not a single ".". InsertCannulaSetupViewController & ReplacePodViewController also will display a confusing "." for certain errors.

For this particular failure, the RileyLink had been damaged and wasn't responding to any commands from Loop. This was generating a PodCommError.commsError error which has no errorDescription string leading to the confusing "." error message being displayed. The commsError errorDescription should be changed from nil to error.localizedDescription so that it produces a useful localized error message for an underlying comms error (both for this particular case and for many other error situations that occur during Loop use). For this particular case doing this would have generated a "RileyLink did not respond in time." error message which would have helped identify that the RileyLInk was indeed not working correctly after taking a dip in the pool... ;-)

itsmojo:
Fixed Thu Sep 17 21:08:24 2020 -0700 with #621

PR #621, opened 2020-09-18
Fix issue #617: OmniKitUI displays a single dot for certain setup errors

itsmojo:
* Return the underlying localized error string for PodCommError.commsError
* Improved lastError handling logic in PairPodSetupViewController,
InsertCannulaSetupViewController & ReplacePodViewController to never display a single
dot (".") for any error and to always display some message for any errors returned.


ps2 (approved):
LGTM!

PR #622, opened 2020-09-18
More pairing fixes  https://github.com/LoopKit/Loop/issues/1338

itsmojo:
* Use an increasing packet # on pairing retries to prevent constant 07 acks
* Automatic retry with new packet # in some cases to prevent a user retry
* Correctly handle pod activation time exceeded faults
* Verify in range RSSI to prevent pods running with malformed pod_addr_1
* Check pod #'s to prevent weird errors with 2 pods pairing

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -1043,7 +1043,7 @@ extension OmnipodPumpManager {
         }
         if self.state.podState?.isFaulted == false && self.state.podState?.unfinalizedBolus?.isFinished == false {
             self.log.info("Skipping Read Pulse Log due to bolus still in progress.")
-            completion(String(describing: PodCommsError.unfinalizedBolus))
+            completion(PodCommsError.unfinalizedBolus.localizedDescription)
             return
         }
 
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -1041,19 +1041,11 @@ extension OmnipodPumpManager {
             completion(PodCommsError.noPodPaired.localizedDescription)
             return
         }
-#if SKIP_ACTIVATION_TIME_EXCEEDED_CHECKING
-        if self.state.podState?.fault == nil && self.state.podState?.unfinalizedBolus?.isFinished == false {
-            self.log.info("Skipping Read Pulse Log due to bolus still in progress.")
-            completion(PodCommsError.unfinalizedBolus.localizedDescription)
-            return
-        }
-#else
         if self.state.podState?.isFaulted == false && self.state.podState?.unfinalizedBolus?.isFinished == false {
             self.log.info("Skipping Read Pulse Log due to bolus still in progress.")
             completion(String(describing: PodCommsError.unfinalizedBolus))
             return
         }
-#endif
 
         let rileyLinkSelector = self.rileyLinkDeviceProvider.firstConnectedDevice
         self.podComms.runSession(withName: "Read Pulse Log", using: rileyLinkSelector) { (result) in
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -1038,11 +1038,19 @@ extension OmnipodPumpManager {
             completion(String(describing: OmnipodPumpManagerError.noPodPaired))
             return
         }
+#if SKIP_ACTIVATION_TIME_EXCEEDED_CHECKING
         if self.state.podState?.fault == nil && self.state.podState?.unfinalizedBolus?.isFinished == false {
             self.log.info("Skipping Read Pulse Log due to bolus still in progress.")
             completion(String(describing: PodCommsError.unfinalizedBolus))
             return
         }
+#else
+        if self.state.podState?.isFaulted == false && self.state.podState?.unfinalizedBolus?.isFinished == false {
+            self.log.info("Skipping Read Pulse Log due to bolus still in progress.")
+            completion(String(describing: PodCommsError.unfinalizedBolus))
+            return
+        }
+#endif
 
         let rileyLinkSelector = self.rileyLinkDeviceProvider.firstConnectedDevice
         self.podComms.runSession(withName: "Read Pulse Log", using: rileyLinkSelector) { (result) in
--- a/OmniKit/PumpManager/PodComms.swift
+++ b/OmniKit/PumpManager/PodComms.swift
@@ -11,6 +11,14 @@ import RileyLinkBLEKit
 import LoopKit
 import os.log
 
+#if !SKIP_RSSI_CHECKS
+fileprivate let enforceRssiLimits = true    // whether to enforce RSSI limit checking
+fileprivate let maxRssiAllowed = 59         // maximum RSSI limit allowed when RSSI limit checking is enabled
+fileprivate let minRssiAllowed = 30         // minimum RSSI limit allowed when RSSI limit checking is enabled
+fileprivate let numRssiRetries = 2          // number of automatic retries if received RSSI value is out of range
+fileprivate let rssiTesting = false         // whether to display message with RSSI value for testing without pairing
+#endif
+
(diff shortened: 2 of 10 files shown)


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1038,11 +1038,19 @@ extension OmnipodPumpManager {
             completion(String(describing: OmnipodPumpManagerError.noPodPaired))
             return
         }
+#if SKIP_ACTIVATION_TIME_EXCEEDED_CHECKING
```
ps2: Please do not use conditional compilation blocks.  There are a lot of things listed in the summary. It would be nice if this was broken up.

ps2 (changes requested):

OmniKit/PumpManager/PodComms.swift
```diff
@@ -11,6 +11,14 @@ import RileyLinkBLEKit
 import LoopKit
 import os.log
 
+#if !SKIP_RSSI_CHECKS
+fileprivate let enforceRssiLimits = true    // whether to enforce RSSI limit checking
+fileprivate let maxRssiAllowed = 59         // maximum RSSI limit allowed when RSSI limit checking is enabled
+fileprivate let minRssiAllowed = 30         // minimum RSSI limit allowed when RSSI limit checking is enabled
+fileprivate let numRssiRetries = 2          // number of automatic retries if received RSSI value is out of range
```
ps2: Good to name these constants, but they should be moved to the local scope they're used in, since they are only used in those places.

OmniKit/PumpManager/PodComms.swift
```diff
@@ -11,6 +11,14 @@ import RileyLinkBLEKit
 import LoopKit
 import os.log
 
+#if !SKIP_RSSI_CHECKS
+fileprivate let enforceRssiLimits = true    // whether to enforce RSSI limit checking
```
ps2: Better to make the decision of whether to include this, rather than adding conditionals and making the code more complex.

OmniKit/PumpManager/PodComms.swift
```diff
@@ -25,6 +31,8 @@ class PodComms: CustomDebugStringConvertible {
 
     public let log = OSLog(category: "PodComms")
 
+    private var startingPacketNumber = 0
```
ps2, line 29: PodComms is an unpersisted object, so any state here will be lost across any app restarts. If the user kills the app mid pairing, or it crashes, will this variable being reset cause any issues?

OmniKit/PumpManager/PodComms.swift
```diff
@@ -41,52 +49,147 @@ class PodComms: CustomDebugStringConvertible {
...
+            if self.podState != nil {
+                self.podState!.messageTransportState = MessageTransportState(packetNumber: transport.packetNumber, messageNumber: transport.messageNumber)
+            } else {
+                self.startingPacketNumber = transport.packetNumber
+            }
+        }
+
+        var didRetry = false
+        var rssiRetries = numRssiRetries
+        while true {
```
ps2, line 84: Not a fan of retrying at this level. It can be *slow* just to do one exchange, and it seems at this level that it's appropriate for the user to have more feedback, and decide what action to take next if a send message fails.

OmniKit/PumpManager/PodComms.swift
```diff
@@ -41,52 +49,147 @@ class PodComms: CustomDebugStringConvertible {
         self.messageLogger = nil
     }
     
-    private func assignAddress(address: UInt32, commandSession: CommandSession) throws -> PodState {
+    // Handles all the common work to send and verify the version response for the two pairing commands, AssignAddress and SetupPod
```
ps2: Could you document which errors get thrown from this?

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -28,6 +28,11 @@ public enum PodCommsError: Error {
     case podSuspended
     case podFault(fault: PodInfoFaultEvent)
     case commsError(error: Error)
+    case podChange
+    case activationTimeExceeded
+    case rssiTooLow
+    case rssiTooHigh
+    case debugFault(str: String)
```
ps2: remove testing code

OmniKit/PumpManager/PodComms.swift
```diff
@@ -11,6 +11,12 @@ import RileyLinkBLEKit
 import LoopKit
 import os.log
 
+fileprivate let enforceRssiLimits = true    // whether to enforce RSSI limit checking
+fileprivate let maxRssiAllowed = 59         // maximum RSSI limit allowed when RSSI limit checking is enabled
+fileprivate let minRssiAllowed = 30         // minimum RSSI limit allowed when RSSI limit checking is enabled
+fileprivate let numRssiRetries = 2          // number of automatic retries if received RSSI value is out of range
+fileprivate let rssiTesting = false         // whether to display message with RSSI value for testing without pairing
```
ps2: remove testing code

ps2 (changes requested):
I have some requests in individual comments. There are a lot of changes here; and of a large enough scope that they can't be considered likely-safe just by code review. After the requested items are addressed, I think this can go in, but I'm leaning heavily on your statements that the logic is well tested.

OmniKit/PumpManager/PodComms.swift
ps2: After discussion, sounds like retrying should be minimal

OmniKit/PumpManager/PodComms.swift
itsmojo: The actual value for this variable is not that important, it has be able to change (not always start at the same packet #) to get the pod out of certain situations where the pod would otherwise be left in a permanent ack state. It does not need to be persistent.

OmniKit/PumpManager/PodComms.swift
itsmojo: Thanks for taking care of this for me. I originally had these declared here as var for testing purposes to allow for easier patching.

OmniKit/PumpManager/PodComms.swift
itsmojo: Thanks for taking care of his for me Pete!

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: Quick note. Having debugFault isn't just for debugging the pairing code itsself (especially now that there will be RSSI limit enforcement that some users might want to tweak), but also for developers needing to be able to test RileyLinks and RileyLink analogues.

OmniKit/PumpManager/PodComms.swift
```diff
@@ -41,52 +44,168 @@ class PodComms: CustomDebugStringConvertible {
...
+                log.error("Received pod response for [lot %u tid %u], expected [lot %u tid %u]", config.lot, config.tid, podState.lot, podState.tid)
+                throw PodCommsError.podChange
+            }
+
+            // Checking RSSI
+            let maxRssiAllowed = 59         // maximum RSSI limit allowed
+            let minRssiAllowed = 30         // minimum RSSI limit allowed
+            if let rssi = config.rssi, let gain = config.gain {
+                let rssiStr = String(format: "Receiver Low Gain: %d.\nReceived Signal Strength Indicator: %d", gain, rssi)
+                log.default("%s", rssiStr)
```
itsmojo, line 139: This snippet should be kept here for any users who might want to customize a different pairing RSSI enforcement policy and for any developers working on RileyLinks and similar devices. We would also need to add back a variable declaration for rssiTesting (or perhaps some alternate name) that can be edited by the user to enable.

                if rssiTesting {
                    throw PodCommsError.debugFault(str: rssiStr)
                }

itsmojo (commented):
I'd like to keep the optional code to be able to repeated print the pairing RSSI values

OmniKit/PumpManager/PodComms.swift
ps2: This is easy enough for someone to add back in. I don't want the code to end up littered with testing conditionals that readers of the code have to bushwack past to try to understand the code.

ps2 (approved):
LGTM!

PR #623, opened 2020-09-18
In-progress boluses and temp-basals should be marked as canceled during deactivation.

itsmojo:
* During deactivation, make sure any in-progress boluses or temp basals are recorded as canceled at the time of deactivation. For faulted pods, or if the cancel program failed, this wasn't happening.
* Do automatic read pulse log as pod messages now persist across pods

### DIFF
--- a/OmniKit/PumpManager/PodCommsSession.swift
+++ b/OmniKit/PumpManager/PodCommsSession.swift
@@ -452,6 +452,42 @@ public class PodCommsSession {
             return DeliveryCommandResult.uncertainFailure(error: error as? PodCommsError ?? PodCommsError.commsError(error: error))
         }
     }
+
+    @discardableResult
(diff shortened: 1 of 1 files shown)


OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -452,6 +452,42 @@ public class PodCommsSession {
             return DeliveryCommandResult.uncertainFailure(error: error as? PodCommsError ?? PodCommsError.commsError(error: error))
         }
     }
+
+    @discardableResult
+    private func handleCancelDosing(deliveryType: CancelDeliveryCommand.DeliveryType, bolusNotDelivered: Double) -> UnfinalizedDose? {
+        var canceledDose: UnfinalizedDose? = nil
+        let now = Date()
+
+        if deliveryType.contains(.basal) {
```
ps2, line 461: This is duplicating code below.

ps2 (changes requested):
Just a small amount of duplicated code; otherwise looks good!

PR #626, opened 2020-09-20
Have Enable/Disable Confirmation Beeps command fail if beepConfig fails

itsmojo:
Add confirmation beeps for the Read Pod Status and Alarms commands


ps2 (approved):
LGTM!

PR #630, opened 2020-09-26
Updated and improved PodInfo subtype names, decoding & use

itsmojo:
* Rename confusing .fault type to a more appropriate .activationTime
* Rename .dataLog type to more appropriate .pulseLogPlus
* Rename .dumpOlderPulseLog to .pulseLogPrevious for consistency
* Always verify the PodInfoResponseSubType value when decoding
* Properly and fully decode PulseLogPlus type
* Add addtional pulse log array size checking when decoding
* Use updated createPulseLog() for decoding all pulse log podInfo types
* Convert PodInfo sub types var variables to let variables
* readPulseLogsRequest() generalized and simplifed to readPodInfo()
* Update readPulseLog() to use new session.readPodInfo() routine
* Change uses of some forced downcasts to conditional downcasts
* When deactivating a faulted pod use new readReadInfo() interface
* When deactivating a faulted pod call updateFromDetailedStatusResponse()
* Use new readReadInfo() interface when deactivating a faulted pod
* Call updateFromDetailedStatusResponse() when deactivating a faulted pod
* Updated and improved PodInfoPulseLog PodInfoTests
* Split cannulaInsertionUnits constant into separate Base and Extra portions
* Change some uses of reservoirVolume to reservoirLevel that were missed


OmniKit/MessageTransport/MessageBlocks/PodInfoActivationTime.swift
```diff
@@ -8,23 +8,26 @@
...
     public let faultEventCode: FaultEventCode
     public let timeActivation: TimeInterval
     public let dateTime: DateComponents
     public let data: Data
     
     public init(encodedData: Data) throws {
         
-        if encodedData.count < 16 {
+        guard let podInfoType = PodInfoResponseSubType.init(rawValue: encodedData[0]), podInfoType == self.podInfoType else {
+            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "activationTime")
```
ps2: typeDescription: "PodInfoResponseSubType"

ps2 (approved):
LGTM. Just a small suggestion.

OmniKit/MessageTransport/MessageBlocks/PodInfoActivationTime.swift
itsmojo: Changed all PodInfoResponseSubType.init() uses to PodInfoResponseSubType()

OmniKit/MessageTransport/MessageBlocks/DetailedStatus.swift
```diff
@@ -35,7 +35,10 @@ public struct DetailedStatus : PodInfo, Equatable {
     
     public init(encodedData: Data) throws {
         
-        if encodedData.count < 21 {
+        guard let podInfoType = PodInfoResponseSubType(rawValue: encodedData[0]), podInfoType == self.podInfoType else {
+            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "detailedStatus")
```
ps2:
```suggestion
            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "PodInfoResponseSubType")
```

OmniKit/MessageTransport/MessageBlocks/PodInfoActivationTime.swift
```diff
@@ -8,23 +8,26 @@
...
     public let faultEventCode: FaultEventCode
     public let timeActivation: TimeInterval
     public let dateTime: DateComponents
     public let data: Data
     
     public init(encodedData: Data) throws {
         
-        if encodedData.count < 16 {
+        guard let podInfoType = PodInfoResponseSubType(rawValue: encodedData[0]), podInfoType == self.podInfoType else {
+            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "activationTime")
```
ps2:
```suggestion
            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "PodInfoResponseSubType")
```

OmniKit/MessageTransport/MessageBlocks/PodInfoConfiguredAlerts.swift
```diff
@@ -32,10 +32,13 @@ public struct PodInfoConfiguredAlerts : PodInfo {
     }
     
     public init(encodedData: Data) throws {
-        if encodedData.count < 11 {
+        guard let podInfoType = PodInfoResponseSubType.init(rawValue: encodedData[0]), podInfoType == self.podInfoType else {
+            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "configuredAlerts")
```
ps2:
```suggestion
            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "PodInfoResponseSubType")
```

OmniKit/MessageTransport/MessageBlocks/PodInfoPulseLog.swift
```diff
@@ -8,68 +8,70 @@
...
+    public let pulseLog      : [UInt32]
     public let data          : Data
-    public let indexLastEntry: UInt16 // the last pulse entry
-    public var pulseLog      : [UInt32]
 
     public init(encodedData: Data) throws {
-        if encodedData.count < 3 || ((encodedData.count - 3) & 0x3) != 0 {
-            throw MessageBlockError.notEnoughData // first 3 bytes missing or non-integral # of log entries
+        guard let podInfoType = PodInfoResponseSubType(rawValue: encodedData[0]), podInfoType == self.podInfoType else {
+            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "pulseLogRecent")
```
ps2:
```suggestion
            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "PodInfoResponseSubType")
```

OmniKit/MessageTransport/MessageBlocks/PodInfoActivationTime.swift
ps2: MessageError.unknownValue is used to report unknown values for raw enum types, and the typeDescription should describe which type the value is unknown for.

I believe this particular value is already checked earlier in the code, so I don't think this will ever be thrown, and could be removed. There is a lot of this checking added in this PR.

There are two things being checked here, in the two clauses of the guard statement: 1) that encodedData[0] is a valid raw value for PodInfoResponseSubType, 2) that the decoded PodInfoResponseSubType is a .activationTime instance. The MessageError.unknownValue error is appropriate for #1, but not #2.  It might be simpler to just do:

```
guard encodedData[0] == self.podInfoType.rawValue else {
```

But even simpler would be to just remove this check (and the similar checks below), as I stated above, as I believe it's already assured earlier in the code.

ps2 (commented):
Renaming and pulse log updates look good. There is some overly defensive code that is a little problematic, though. I think it can just be removed; fixing it is also a little awkward.

OmniKit/MessageTransport/MessageBlocks/DetailedStatus.swift
itsmojo: Done

OmniKit/MessageTransport/MessageBlocks/DetailedStatus.swift
```diff
@@ -34,9 +34,8 @@ public struct DetailedStatus : PodInfo, Equatable {
     public let data: Data
     
     public init(encodedData: Data) throws {
-        
-        guard let podInfoType = PodInfoResponseSubType(rawValue: encodedData[0]), podInfoType == self.podInfoType else {
-            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "detailedStatus")
+        guard encodedData[0] == self.podInfoType.rawValue else {
+            throw MessageError.unknownValue(value: encodedData[0], typeDescription: "PodInfoResponseSubType")
```
ps2: This is no longer an unknown value for PodInfoResponseSubType, so I don't think unknownValue is appropriate here.  The error is that you're expecting a podInfoType for [0] and getting something else. I don't see an appropriate MessageError type for this; might need to make a new one.  But as I stated before, I think you can just remove this check, no? We only get into this code if we've already checked that byte.

Same for the similar blocks below.

ps2: Doing an assertionFailure() would be another way of keeping an assertion here for safety during development.

itsmojo: The case I wanted to protect against where the podInfoType could be wrong would be a downcast to the wrong podInfo type such as in:
```quote
    let podInfoResponse: PodInfoResponse = try send([GetStatusCommand(podInfoType: .pulseLogRecent)])
    if let podInfoPulseLogPrevious = podInfoResponse.podInfo as? PodInfoPulseLogPrevious
```
I think using asserts would be the right way to go for this stuff, but I didn’t see them used that often in Loop and so I thought that its here might be frowned on.

ps2: At the point of the downcast on this line, we have already checked that 1) the podInfoResponseSubType is a valid number, and 2) that the message has decoded successfully as that type.  That happens in the `public init(encodedData: Data)` constructor for PodInfoResponse. So at this point in the code, no matter what is returned from the pod, the parsing will not call a constructor for one of the PodInfo subtypes with the wrong value in byte[0].  The downcast will succeed or result in a null depending on whether the subtype was a PulseLogPrevious type.

itsmojo: Got it. The self.podInfoType checking was only verifying that these values were initialized correctly in the source files. I have now removed these checks.

PR #631, opened 2020-09-29
Fix date picker appearance in iOS 14

novalegra:
This is a small fix to address the date picker UI change that was made in iOS 14; I added a change to force the picker to show up in the `.wheels` style that it had before.


ps2:
Thanks Anna!

PR #633, opened 2020-10-01
Updated pod fault handling and DetailedStatus decoding

itsmojo:
* Add new detailedStatus isFaulted var
* Updated and unified pod fault handling
* Updated getDetailedStatus() to handle pod faults
* Reimplemented DetailedStatus VV and 0X bytes as optionals
* Created new ErrorEventInfo type for the optional VV byte
* Deleted incorrect & unneeded OmniKit/Model/LogEventErrorCode.swift


ps2 (approved):
LGTM! Thanks Joe!

PR #634, opened 2020-10-03
Allow recovery from incomplete pod pairing

ericmason:
If the pod doesn't finish pairing for some reason, for example you get stuck in `setupProgress: cannulaInserting`, you will end up with a screen like this with no way to finish the setup or unpair the pod:
120832805_10224525845514649_4239984110674334030_n

This PR includes the configuration section on the Pod Settings page in that scenario so that you can finish setting up the pod.


ps2:
Dup of https://github.com/ps2/rileylink_ios/pull/632

PR #635, opened 2020-10-07
Two minor Read Pod Status display tweaks

itsmojo:
Display the pod Pulse Count as just a decimal #
For no alerts display "No alerts" instead of an empty string

### DIFF
--- a/OmniKitUI/ViewControllers/CommandResponseViewController.swift
+++ b/OmniKitUI/ViewControllers/CommandResponseViewController.swift
@@ -59,7 +59,7 @@ extension CommandResponseViewController {
 
         result += String(format: LocalizedString("Delivery Status: %1$@\n", comment: "The format string for Delivery Status: (1: delivery status string)"), String(describing: status.deliveryStatus))
 
-        result += String(format: LocalizedString("Pulses (incl. prime & insert): %1$@ U\n", comment: "The format string for total insulin pulses (1: total pulses in units)"), status.totalInsulinDelivered.twoDecimals)
+        result += String(format: LocalizedString("Pulse Count: %1$d\n", comment: "The format string for Pulse Count (1: pulse count)"), Int(status.totalInsulinDelivered / Pod.pulseSize))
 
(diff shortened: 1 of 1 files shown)


OmniKitUI/ViewControllers/CommandResponseViewController.swift
```diff
@@ -59,7 +59,7 @@ extension CommandResponseViewController {
 
         result += String(format: LocalizedString("Delivery Status: %1$@\n", comment: "The format string for Delivery Status: (1: delivery status string)"), String(describing: status.deliveryStatus))
 
-        result += String(format: LocalizedString("Pulses (incl. prime & insert): %1$@ U\n", comment: "The format string for total insulin pulses (1: total pulses in units)"), status.totalInsulinDelivered.twoDecimals)
+        result += String(format: LocalizedString("Pulse Count: %1$d\n", comment: "The format string for Pulse Count (1: pulse count)"), Int(status.totalInsulinDelivered / Pod.pulseSize))
```
ps2, line 62: Might make some sense for DetailedStatus to just store pulses, and it can provide that in units of insulin with a helper, so we don't have to do back & forth math here.

itsmojo: I had considered doing this, but decided against this approach as it could open up a can of worms.  DetailedStatus.totalInsulinDelivered itself is actually used as an insulin amount elsewhere (updateFromDetailedStatusResponse() & PodInsulinMeasurements()) and there is an analogous value for the exact same pod variable that is returned in the standard StatusResponse (but there it is called "insulin" instead of "totalInsulinDelivered") that is always used as an insulin value throughout Loop. If we converted DetailedStatus.totalInsulinDelivered to a pulse count to save the single conversion here, have we'd have to do a pulse count to insulin amount conversion in multiple places and we'd an inconsistent field for what should be the exact same variable in DetailedStatus & StatusResponse (unless we updated StatusResponse to be consistent for the same pod value in which case there would need to be even pulses to insulin amount conversions). Additionally there are a couple of other values reported by the pod as a pulse count bolusNotDelivered & reservoirLevel) for both DetailedStatus and StatusResponse that would either be inconsistent or require even more pulse count to insulin amount conversions.

PR #636, opened 2020-10-13
Have diagnostic commands display complete error messages

itsmojo:
* Have Read Pulse Log only read last 50 pulse log entries
* Change readPulseLog() to have separate success and failure responses
* Check for bolus in progress before running Read Pod Status
* Change default let mockCommsErrorDuringPairing value to false
* Improved Play Test Beeps success message text
* Generalize twoDecimals for any Double w/o a reservoirLevel reference

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -950,7 +950,8 @@ extension OmnipodPumpManager {
             completion(.failure(PodCommsError.noPodPaired))
             return
         }
-        if self.state.podState?.isFaulted == false && self.state.podState?.unfinalizedBolus?.isFinished == false {
+        guard state.podState?.isFaulted == true || state.podState?.unfinalizedBolus?.scheduledCertainty == .uncertain || state.podState?.unfinalizedBolus?.isFinished != false else
+        {
             self.log.info("Skipping Read Pod Status due to bolus still in progress.")
             completion(.failure(PodCommsError.unfinalizedBolus))
             return
@@ -1005,7 +1006,7 @@ extension OmnipodPumpManager {
             completion(OmnipodPumpManagerError.noPodPaired)
             return
         }
-        guard self.state.podState?.unfinalizedBolus?.isFinished != false else {
+        guard state.podState?.unfinalizedBolus?.scheduledCertainty == .uncertain || state.podState?.unfinalizedBolus?.isFinished != false else {
             self.log.info("Skipping Play Test Beeps due to bolus still in progress.")
             completion(PodCommsError.unfinalizedBolus)
             return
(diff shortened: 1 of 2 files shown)


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -950,6 +950,12 @@ extension OmnipodPumpManager {
             completion(.failure(PodCommsError.noPodPaired))
             return
         }
+        guard state.podState?.isFaulted == true || state.podState?.unfinalizedBolus?.scheduledCertainty == .uncertain || state.podState?.unfinalizedBolus?.isFinished != false else
```
ps2: I don't think we need to prevent status when a bolus is running, or when there is an uncertain bolus. Does the status command not work when a pod is faulted?

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1027,15 +1033,16 @@ extension OmnipodPumpManager {
         }
     }
 
-    public func readPulseLog(completion: @escaping (String) -> Void) {
+    public func readPulseLog(completion: @escaping (Result<String, Error>) -> Void) {
```
ps2, line 1030: 👍

ps2 (changes requested):
Generally looks good, but I don't think we need to prevent status reading in a couple of conditions that this is adds checking for.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: OmnipodPumpManager::getPodStatus() has never allowed a getStatus() during a certain in-progress bolus (which matches the PDM behavior which only allows for a cancel command during a bolus). Now that readPodStatus() is implemented using getDetailedStatus() (which has a 2 packet response instead of a single packet response), it would be prudent to have the same enforcement in readPodStatus() that we've always had in getPodStatus() to avoid pod comms during certain in-progress boluses when possible. I recently had a $34 fault (reset from unknown cause) on the 3rd day of use for my own personal use (which I haven't had for some time) after doing a number of the detailed "Read Pod Status" commands (both w/ and w/o active boluses) which was part of my inspiration for adding this check.

getDetailedStatus() will succeed on a faulted pod while getStatus() (what used was formally used by readPodStatus()) will not succeed. The isFaulted test in this guard simply allows Loop to fetch the DetailedStatus for a pod that has faulted mid bolus (i.e., a bolus is never actually in-progress for a faulted pod even though unfinalizedBolus might indicate that it is).

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: I thought you'd like this change... ;-)

OmniKit/PumpManager/OmnipodPumpManager.swift
ps2: I'm just asking you to remove the code here that prevents the pod status page from fetching fresh information while bolusing.

itsmojo: Unless I am misreading something, this check is only for the heavier-weight readPodStatus() function which has to be manually invoked by the user using "Read Pod Status" and doesn't change a thing regarding pod status page refreshes. The refreshing of the pod status information uses the lighter-weight (but similar sounding) getPodStatus() function which didn't get changed in this PR which has always checked for a bolus in progress.

itsmojo: Reluctantly removed the bolus checking in readPodStatus() so that hopefully this PR can land

PR #638, opened 2020-10-17
Rework to avoid having to do a get status before each bolus

itsmojo:


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1364,32 +1364,64 @@ extension OmnipodPumpManager: PumpManager {
...
                     completion(.failure(PumpManagerError.communication(error as? LocalizedError)))
                     return
                 }
             }
 
-            guard !podStatus.deliveryStatus.bolusing else {
-                completion(.failure(PumpManagerError.communication(PodCommsError.unfinalizedBolus)))
+            var doGetStatus = false
+            var doFinalizeFinishedDoses = false
+            let preError = self.setStateWithResult({ (state) -> PodCommsError? in
```
ps2: It looks like this block is not actually updating state; you can just access state, or make a local let of state.

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1364,32 +1364,64 @@ extension OmnipodPumpManager: PumpManager {
...
                 } catch let error {
+                    self.log.error("enactBolus: error resuming suspended pod: %s", String(describing: error))
                     completion(.failure(PumpManagerError.communication(error as? LocalizedError)))
                     return
                 }
             }
 
-            guard !podStatus.deliveryStatus.bolusing else {
-                completion(.failure(PumpManagerError.communication(PodCommsError.unfinalizedBolus)))
+            var doGetStatus = false
```
ps2: Reading below, it looks like this `doGetStatus` is set when you want to attempt an uncertain bolus reconciliation?  If so, maybe we can give it a more descriptive name, like `attemptUncertainBolusReconciliation`, or make a comment; this method is getting fairly hairy and hard to read.

OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -464,6 +465,7 @@ public class PodCommsSession {
                 podState.unfinalizedBolus = UnfinalizedDose(bolusAmount: units, startTime: Date(), scheduledCertainty: .uncertain)
                 return DeliveryCommandResult.uncertainFailure(error: podCommsError)
             }
+            podState.updateFromStatusResponse(status)
```
ps2: getStatus() already does an `updateFromStatusResponse`

OmniKit/PumpManager/PodState.swift
```diff
@@ -196,10 +196,14 @@ public struct PodState: RawRepresentable, Equatable, CustomDebugStringConvertibl
         }
     }
     
-    private mutating func updateDeliveryStatus(deliveryStatus: DeliveryStatus) {
+    private mutating func updateDeliveryStatus(deliveryStatus: DeliveryStatus, podProgressStatus: PodProgressStatus, bolusNotDelivered: Double) {
         finalizeFinishedDoses()
 
-        if let bolus = unfinalizedBolus, bolus.scheduledCertainty == .uncertain {
+        // Add the unfinalizedBolus here if there isn't one and are currently bolusing in a ready state.
+        if unfinalizedBolus == nil && deliveryStatus.bolusing && podProgressStatus.readyForDelivery {
```
ps2: Have you actually observed a state like this outside of killing Xcode?

ps2 (commented):
I think the idea behind this is good; a few suggestions, and the attempt to partially recover an unrecorded bolus sounds like a separate thing.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: This is the same basic logic that is used in enactTempBasal() which doesn't allow for a new TBR to be sent if the pod is suspended or an unfinalized bolus is currently in progress. If enactTempBasal and enactBolus() can both be reworked to use a simpler style for the preError checking, please to go for it -- but I'm not comfortable doing this myself at this time. Additionally I know that it has had months of testing in its current form.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: Yes, Loop still needs to the get status before a bolus to attempt an uncertain bolus reconciliation as the self.log.info("enactBolus: doing getStatus with uncertain bolus scheduled certainty") call right before setting doGetStatus documented in lieu of a comment. Nevertheless, I'll rename this variable to `attemptUncertainBolusReconciliation` to be even more explicit about the reason for needing to do the get status for this case.

OmniKit/PumpManager/PodState.swift
itsmojo: No, but having like this is critical for this PR. If a pod ever gets a bolus command when a bolus is currently in progress, the pod will immediately fail with an $31 fault. The unfinalizedBolus is what keeps track of in-progress boluses, so it's critical to have this correctly set up for any in-progress bolus. I used an Xcode breakpoint to simulate a user stopping Loop at the exact wrong time right after the bolus has started, but before the unfinalizedBolus is created. Without having this code, Loop would fault the pod if an automatic bolus was attempted on Loop restart if the bolus was still on going if the app was stopped at the exact spot. Having this code also protects against a pod fault if Loop ever has some logic error where an unfinializedBolus is never initialized or it is incorrectly removed/deleted before the bolus has actually completed.

In summary, implementing a partial recovery of an unrecorded/lost bolus isn't actually a separate thing, but it is required to avoid $31 pod faults when not doing a get status call before each bolus.

OmniKit/PumpManager/PodCommsSession.swift
itsmojo: Ah yes, I messed this. The getStatus() done in the attempt to verify an uncertain bolus will indeed handle the updateFromStatus() so this one is indeed redundant.

OmniKit/PumpManager/OmnipodPumpManager.swift
ps2: You can keep the same style as a function if you want:
```
let preError = {
    if case .some(.suspended) = state.podState?.suspendState {
        self.log.error("enactBolus: failing with suspended pod")
        return .podSuspended
    }
    ...
}()

The setStateWithResult makes it looks like it's updating state, which it's not.

File: OmniKit/PumpManager/OmnipodPumpManager.swift [hunk 5]
```diff
@@ -1364,32 +1364,63 @@ extension OmnipodPumpManager: PumpManager {
...
                     return
                 }
             }
 
-            guard !podStatus.deliveryStatus.bolusing else {
-                completion(.failure(PumpManagerError.communication(PodCommsError.unfinalizedBolus)))
+            var attemptUncertainBolusReconciliation = false
+            var finalizeFinishedDoses = false
+            let preError = self.setStateWithResult({ (state) -> PodCommsError? in
+                if case .some(.suspended) = state.podState?.suspendState {
```
ps2: Above, we just checked the same thing, and attempt to resume the pod if it's suspended. If the resume fails, we error and return. This is redundant.

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1364,32 +1364,63 @@ extension OmnipodPumpManager: PumpManager {
...
+                    self.log.error("enactBolus: failing with suspended pod")
+                    return .podSuspended
+                }
+                if let unfinalizedBolus = state.podState?.unfinalizedBolus {
+                    if unfinalizedBolus.scheduledCertainty == .uncertain {
+                        self.log.info("enactBolus: doing getStatus with uncertain bolus scheduled certainty")
+                        attemptUncertainBolusReconciliation = true
+                        return nil
+                    }
+                    if !unfinalizedBolus.isFinished {
```
ps2: This check does not guarantee the pod is actually done bolusing. It's an estimate. If our estimated time to complete the bolus is too short, and the pod is still bolusing, then we could trigger a fault, no?

OmniKit/PumpManager/PodState.swift
ps2: This seems like unlikely to be used code, that makes understanding what's going on here more difficult than it needs to be. I really dislike this pattern of defensive coding for problems that haven't been observed and are very unlikely to ever happen. It seems a much more real possibility that the pod might take longer to deliver a bolus than we expect (based on clock speed differences, or variability in memory wire action timing, which the pod observes), which is why we had the status check in the first place.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: Rewrote enactBolus(), enactTempBasal() and setTime() to avoid using setStateWithResult

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: This is correct but has never happened in all its previous use. Reworked logic to use new isBolusPositivelyFinished variable that adds 20% more time as pod will fault if any pulse takes more than 20% longer than nominal to complete. If the isFinished is true but isBolusPositivelyFinished is false, then a getStatus call will be used to verify if the bolus is indeed completed to avoid any possible pod fault.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: This was indeed paranoid code which also allowed the early code to be completed removed if not desired. It has been removed.

OmniKit/PumpManager/PodState.swift
itsmojo: Rewrote things a bit to simplify this part. This code is needed to handle a bolus being delivered and then the user moving out of range (before the response is received & the subsequent getStatus completes trying to verify the bolus) and then the user stops Loop. There is new logic in enactBolus() to do a getStatus() if enough time has passed for the bolus to finish, but not enough time for Loop to be absolutely positive.

PR #639, opened 2020-10-20
Have a user requested refresh emit pod confirmation beeps when enabled

itsmojo:


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -722,18 +722,18 @@ extension OmnipodPumpManager {
         }
     }
 
-    public func refreshStatus(completion: ((_ result: PumpManagerResult<StatusResponse>) -> Void)? = nil) {
+    public func refreshStatus(userRequest: Bool, completion: ((_ result: PumpManagerResult<StatusResponse>) -> Void)? = nil) {
```
ps2: Rather than making PumpManager aware of the use case, I'd rather expose the meaning of what does to the caller, so `userRequest` -> `emitConfirmationBeeps`.  Logic doesn't need to change, but it lets the caller decide when it's appropriate to beep.

ps2 (changes requested):

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: Renamed userRequest parameter to expose the meaning as requested

PR #641, opened 2020-10-26
Verify pod setup completion to fix LookKit issue #1369

itsmojo:
Don't have OmnipodPumpManager check cannula insertion as errors are lost
Have InsertCannulaSetupViewController do completion checking & report errors
Add optional finish case & Finish button to InsertCannulaSetupViewController
Add additional guard to prevent potential bad setup state problem
Remove unused cancelErrorCount var in InsertCannulaSetupViewController

### DIFF
--- a/OmniKitUI/ViewControllers/InsertCannulaSetupViewController.swift
+++ b/OmniKitUI/ViewControllers/InsertCannulaSetupViewController.swift
@@ -33,8 +33,6 @@ class InsertCannulaSetupViewController: SetupTableViewController {
         }
     }
     
-    private var cancelErrorCount = 0
-    
     override func viewDidLoad() {
         super.viewDidLoad()
         
@@ -61,6 +59,7 @@ class InsertCannulaSetupViewController: SetupTableViewController {
         case initial
         case startingInsertion
         case inserting(finishTime: CFTimeInterval)
+        case finish
         case fault
         case ready
     }
@@ -71,7 +70,7 @@ class InsertCannulaSetupViewController: SetupTableViewController {
             case .initial:
                 activityIndicator.state = .hidden
                 footerView.primaryButton.isEnabled = true
-                footerView.primaryButton.setConnectTitle()
+                footerView.primaryButton.setInsertCannulaTitle()
             case .startingInsertion:
                 activityIndicator.state = .indeterminantProgress
                 footerView.primaryButton.isEnabled = false
@@ -80,6 +79,10 @@ class InsertCannulaSetupViewController: SetupTableViewController {
                 activityIndicator.state = .timedProgress(finishTime: CACurrentMediaTime() + finishTime)
                 footerView.primaryButton.isEnabled = false
                 lastError = nil
+            case .finish:
+                activityIndicator.state = .hidden
+                footerView.primaryButton.isEnabled = true
+                footerView.primaryButton.setFinishTitle()
             case .fault:
                 activityIndicator.state = .hidden
                 footerView.primaryButton.isEnabled = true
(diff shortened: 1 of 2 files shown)


OmniKitUI/ViewControllers/InsertCannulaSetupViewController.swift
```diff
@@ -115,20 +118,44 @@ class InsertCannulaSetupViewController: SetupTableViewController {
...
             } else if lastError != nil {
-                continueState = .initial
+                continueState = initialOrFinish
             }
         }
     }
 
+    // true if pod setup progress has been verified as completed
+    private var setupIsCompleted: Bool {
+        if pumpManager.state.podState?.setupProgress == .completed {
```
ps2: [nit] This can just be `return pumpManager.state.podState?.setupProgress == .completed`

OmniKitUI/ViewControllers/InsertCannulaSetupViewController.swift
```diff
@@ -61,6 +59,7 @@ class InsertCannulaSetupViewController: SetupTableViewController {
         case initial
         case startingInsertion
         case inserting(finishTime: CFTimeInterval)
+        case finish
```
ps2: Maybe `finish` -> `needsAdditionalCheck`? I'm having some cognitive dissonance about "ready" being the final state and "finish" being an earlier state. Maybe `ready` changes to `finishedInsertion`?

ps2 (approved):
I didn't test this, but it reads correct to me; just some comments on naming and a nit.

ps2:
I'm wondering if the "finish" state would be better communicated to the user with some messaging like "Could not confirm successful insertion with the pod. Tap "Confirm Insertion" to check again". And change the button to "Confirm Insertion".

OmniKitUI/ViewControllers/InsertCannulaSetupViewController.swift
itsmojo: Sounds good. Will do.

OmniKitUI/ViewControllers/InsertCannulaSetupViewController.swift
itsmojo: Loop has always used `ready` for the final state in all the pod setup views (Pairing, Insert Cannula & Replace) and this does seem to succinctly and consistently describe this state, so I believe that continuing to use `ready` here for this seems reasonable. Perhaps it's just a matter to find a better name for the new state if there is some cognitive dissonance, even though I'm not actually sure what might be causing this when using `finish` for a new earlier state. I could more easily understand if the name and button label were `finished` and "Finished" which imply the current task has already been completed instead of `finish` and "Finish" which imply that action is needed to complete the current task (of inserting the cannula) which would seem to make sense.

Nevertheless, I could go with something like `needsInsertionCheck` for this new state with a button label of "Check Insertion" as it sort of matches the name of the routine that will be used to do the confirmation, checkCannulaInsertionFinished() even though this really isn't what is happening.

itsmojo:
I'd rather not display anything saying "Could not confirm successful insertion with the pod" on errors as as the only cannula insertion confirmation that can be done is for the user to manually look for the pink color in the viewing window on top the pod. Also, by using the normal Loop error message for whatever error might have occurred, the user gets more information about what actually went wrong and how to correct it (e.g., "No response from pod. Please try repositioning the pod or the RileyLink and try again."). Additionally, using this approach there will be no additional new string to have to translate... ;-)

itsmojo:
Decided to go with the more explicit `needsCheckInsertion` for the new State enum name and "Recheck Cannula Insertion" for the button label. Using "Recheck" in the button label description instead of "Check" helps give context for what the displayed error was about and is more correct & using "Cannula Insertion" instead of just "Insertion" matches the View title better.

ps2:
Thanks!

PR #643, opened 2020-11-29
Use  a beep message block for confirmation beeps instead of a separate message

itsmojo:
Remove now unneeded workaround to handle cancel all confirmation beep
New confirmation beep sequence for the Read Pulse Log command
Remove tempBasalConfirmationBeeps debug support as directed :(


ps2 (approved):
LGTM! Thanks Joe.

PR #644, opened 2020-11-30
Avoid reading past A17 and A21 alarms

ps2:
When Medtronic pumps lose track of time during battery changes, it can interfere with reconstruction of actual event time for pump history, often causing events to appear with far future dates. This change stops processing of events past the device reset alarms A17 and A21, that signify a pump time chang may have occured without corresponding "Change Time" events.


RedCyclist1:
Pete, 
I have this situation and been off Loop for 10 days trying to troubleshoot. I have updated everything, wiped my phone, updated computer, phone and watch... xcode and rebuilt loop, but I have not fixed this issue with my pump... can we schedule some time to fix?

Kryptonit3:
> Pete,
> I have this situation and been off Loop for 10 days trying to troubleshoot. I have updated everything, wiped my phone, updated computer, phone and watch... xcode and rebuilt loop, but I have not fixed this issue with my pump... can we schedule some time to fix?

This fix looks like it was merged into dev and not master. May need to manually switch the branch when pulling Loop. Are you using automatic-bolus, or workspace or master? Looks like the Cartfile for AB branch is pulling `github "ps2/rileylink_ios" "carthage-pin"` which doesn't have this fix yet if that is where Loop gets it's reference from. There is a way to manually apply a pull request as a patch but I have no idea how.

EDIT

Here is the patch URL https://github.com/ps2/rileylink_ios/pull/644.patch not sure how to apply it after downloading Loop and clicking build.

Kryptonit3:
@ps2  was this fix removed from the dev branch? Any idea when it will be included with the branch that Loop AB uses by default (carthage-pin)? It is still very much an app breaking issue.

PR #647, opened 2021-01-07
Add toggle to optionally disable mySentry

elnjensen:
This code adds a setting to the Pump Settings menu for Medtronic x23/x54 pumps that allows the user to disable the use of MySentry, potentially improving battery life on the RileyLink. 

- Should there be additional logging somewhere to track any change in mySentry settings? 
- Does more need to be done when toggling besides changing `idleListeningState`? 

Still to do: adding a dialog page during the pump setup process to explain and initialize the setting.

### DIFF
--- a/MinimedKitUI/MinimedPumpSettingsViewController.swift
+++ b/MinimedKitUI/MinimedPumpSettingsViewController.swift
@@ -193,11 +193,7 @@ class MinimedPumpSettingsViewController: RileyLinkSettingsViewController {
                 cell.detailTextLabel?.text = String(describing: pumpManager.preferredInsulinDataSource)
             case .useMySentry:
                 cell.textLabel?.text = LocalizedString("Use MySentry", comment: "The title text for the preferred MySentry setting config")
-                if pumpManager.state.pumpModel.hasMySentry {
-                    cell.detailTextLabel?.text = pumpManager.useMySentry ? "Yes" : "No"
-                } else {
-                    cell.detailTextLabel?.text = "N/A"
-                }
+                cell.detailTextLabel?.text = pumpManager.useMySentry ? "Yes" : "No"
             case .timeZoneOffset:
                 cell.textLabel?.text = LocalizedString("Change Time Zone", comment: "The title of the command to change pump time zone")
 
--- a/MinimedKitUI/MinimedPumpSettingsViewController.swift
+++ b/MinimedKitUI/MinimedPumpSettingsViewController.swift
@@ -101,6 +101,7 @@ class MinimedPumpSettingsViewController: RileyLinkSettingsViewController {
         case batteryChemistry
         case preferredInsulinDataSource
         case insulinType
+        case useMySentry
     }
 
     // MARK: UITableViewDataSource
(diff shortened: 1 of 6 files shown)


MinimedKitUI/MinimedPumpSettingsViewController.swift
```diff
@@ -187,6 +190,13 @@ class MinimedPumpSettingsViewController: RileyLinkSettingsViewController {
             case .preferredInsulinDataSource:
                 cell.textLabel?.text = LocalizedString("Preferred Data Source", comment: "The title text for the preferred insulin data source config")
                 cell.detailTextLabel?.text = String(describing: pumpManager.preferredInsulinDataSource)
+            case .useMySentry:
+                cell.textLabel?.text = LocalizedString("Use MySentry", comment: "The title text for the preferred MySentry setting config")
+                if pumpManager.state.pumpModel.hasMySentry {
```
ps2: This line will never evaluate to false, right? Pumps that do not have mySentry will not have this row. This conditional can be removed, I think.

RileyLinkKit/PumpOpsSession.swift
```diff
@@ -65,7 +65,7 @@ extension PumpOpsSession {
 
         let shortPowerMessage = PumpMessage(settings: settings, type: .powerOn)
 
-        if pump.pumpModel == nil || !pump.pumpModel!.hasMySentry {
+        if pump.pumpModel == nil || !(pump.pumpModel!.hasMySentry && pump.useMySentry) {
```
ps2: I'm not 100% sure, but I think the wakeup requirements vary based on the model number of the pump, and not whether mysentry is enabled, so I think this might be adding a slower wakeup for pumps that do not need it.

ps2 (commented):
This looks good. Just a couple comments, and a possible larger improvement: it probably makes sense to ask a first time user setting up the pump if they want to use mysentry or not, and to explain (succinctly) why they may want to enable or disable this. A page in the pump setup flow would achieve this nicely. I'd prefer that we add that before accepting this PR, but it proves problematic to do so, let me know.

ps2:
@elnjensen I had some comments on some of the code. And it looks like there's conflicts now.  Could you update this to current dev, and maybe remove the unneeded conditional?

elnjensen:
Sorry for the slow response - have been meaning to get back to this.  My 523 pump died but thanks to Jeremy I now have a 723 I can use for testing.  I’ll clean things up w.r.t. current dev and also address the conditional.  Then I’ll take a look at the pump setup screens and see what I can do there.  

What’s your thinking on your comment about the code change in the pump wakeup part - can that definitely come out?  If so it actually simplifies the code changes a bit since then the mySentry flag doesn’t need to be in pumpState, just in pumpManagerState.

MinimedKitUI/MinimedPumpSettingsViewController.swift
elnjensen: Done, thanks.

RileyLinkKit/PumpOpsSession.swift
elnjensen: Reverted.

PR #652, opened 2021-02-01
Additional Omnipod automatic bolus support

itsmojo:
Add an alternate Program Reminder value for automatic boluses
Add optional acknowledgement beep for automatic boluses
Add new persistent automaticBolusBeeps state for future UI use
Add a nbsp for the type 11 PDM Ref string for consistency


ps2:
Thanks!

PR #654, opened 2021-02-04
Support for suspend reminder and suspend time expired alerts

itsmojo:


OmniKit/PumpManager/PodCommsSession.swift
```diff
@@ -542,6 +542,66 @@ public class PodCommsSession {
...
+            return CancelDeliveryResult.certainFailure(error: PodCommsError.nonceResyncFailed)
+        } catch PodCommsError.rejectedMessage(let errorCode) {
+            return CancelDeliveryResult.certainFailure(error: PodCommsError.rejectedMessage(errorCode: errorCode))
+        } catch let error {
+            podState.unfinalizedSuspend = UnfinalizedDose(suspendStartTime: Date(), scheduledCertainty: .uncertain)
+            return CancelDeliveryResult.uncertainFailure(error: error as? PodCommsError ?? PodCommsError.commsError(error: error))
+        }
+    }
+
+    // Cancels any suspend related alerts, should be called when resuming after using suspendDelivery()
```
ps2, line 591: Is this how the PDM does it?  With two separate messages, one for the start program and one for alert config? Seems like potential for weird issues where the resume succeeds, but the alert reconfig fails, and alerts continue.

itsmojo: This is indeed how the PDM does it. Since the pod will not allow any other message to be combined with any insulin command, the controlling software must do this with a separate $19 Configure Alerts message to cancel both the suspend related alerts right after the $1A set basal schedule command message. However, the PDM sets the high bit in the B9 bytes for the $1A message which will cause the pod to fault if it doesn't receive a non-status request command in the next 4 minutes. This works for the PDM because it probably will not allow a cancel between the two commands and will keep insisting on retrying if the pod is out of range after the first command is sent. I think that with Loop we should not use the send() expectFollowOnMessage option for resumes at this time because we'd have to rework too much to make it sensible to use and the value of using this option seems pretty small. At the worse case if the Loop user cancels a resume operation between the $1A & $19 commands (or the single packet $19 command message doesn't make it to the pod), the pod will not fault but just continue with the pod suspend alerts until the user manually acknowledges the alert to clear the beeping.

We could consider setting the high B9 bit for the initial basal programming when pairing (i.e., using the expectFollowOnMessage send() option in programInitialBasalSchedule()) as the PDM does, but we'd have to make sure that the user can't cancel the pairing operation between the session.programInitialBasalSchedule() and the session.insertCannula() calls done by OmnipodPumpManager:insertCannula(). This is something to consider doing in some other PR, but again I don't think using this option adds much value as I believe that Loop should be able to recover without its use.

PR #655, opened 2021-02-07
Verify and use pod constants returned in the SetupPod response

itsmojo:
Enforce 0.05U pulse size must mandatorily match as per PDM
Verify secondsPerBolusPulse & secondsPerPrimePulse are 2 & 1
Verify that reported pod lifetime matches expected 80 hours
Adjust prime and cannula insertion bolus sizes to pod values
Updated code & comments for setup bolusing to correctly scale
Updated comments for pulse unit sizing is for U100 insulin
Update pairing RSSI message and fix a string format error
Fix to properly handle activation timeout error w/o a pod fault


OmniKit/Model/Pod.swift
```diff
@@ -57,13 +57,15 @@ public struct Pod {
     // Minimum duration of a single basal schedule entry
     public static let minimumBasalScheduleEntryDuration = TimeInterval.minutes(30)
 
-    // Amount of insulin delivered with 1 second between pulses for priming
-    public static let primeUnits = 2.6
+    // Default amount for priming bolus using secondsPerPrimePulse timing.
+    // Can potentially be overrided with value returned by pod during the pairing process.
+    public static var primeUnits = 2.6
```
ps2: These should be constants. There should be no modifiable state here.

ps2 (changes requested):
Thanks for the updated decoding!  I think generally this looks good, except for the storing of state on the Pod struct. If we really need to track this as state (not sure that's really needed), then it should go on the PodState or PumpManager State.

OmniKit/Model/Pod.swift
itsmojo: Changed

Issue #656, opened 2021-03-01
Active Time display doesn't account for clock differences and can be inconsistent


itsmojo:
The Pod Setting's "Active Time" display isn't actually based on the timeActive value returned from pod on each response and thus can show inconsistent values. Due to the pod's internal clock inaccuracies, the displayed value can be different from the "Read Pod Status" command's "Pod Active" display (which actually displays the actual pod reported timeActive value) and can disagree with the "Expires" / "Expired" display (which is updated dynamically based on each pod's timeActive value returned) as shown in the following example.

![IMG_4096](https://user-images.githubusercontent.com/23020111/109466609-9b26e080-7a2f-11eb-99bc-dd42676570c1.PNG)
![IMG_4097](https://user-images.githubusercontent.com/23020111/109466625-a1b55800-7a2f-11eb-8017-eb579f6ebfa6.PNG)

The Read Pod Status "Pod Active" display shows that the pod is reporting that it has been active for 3d, 1h, 3m which disagrees with the Pod Setting's "Active Time" display that says the pod's active life was only 3d, 0h, and 59m. Also the Pod Setting's "Expired" display indicates that the pod expired over a hour ago at the time these screenshots were taken (and which was confirmed by the 1 hour "Expiration advisory" beep that had just occurred a few minutes prior), but the "Active Time" value of 3d, 0h, 59m state that an hour has not yet passed since the pod nominal 72 hour expiration. So in this example, the pod's clock was about 4 minutes too fast over 3 days (which is pretty typical value) and the "Active Time" display is disagreeing with two other displayed values as well as the expiration advisory alarms that are all based on the pod's notion of time.

A better behavior would to simply compute the displayed "Active Time" using the dynamically adjusted podState.expiresAt variable (which already accounts for pod clock differences) instead using the fixed and never adjusted podState.activatedAt variable or to have Loop save & display the actual timeActive value as returned from the pod in each response as a new persistent podState state variable.

ps2:
Displaying actual activated at time resolves this: https://github.com/ps2/rileylink_ios/pull/677

PR #657, opened 2021-03-02
Fix issue #656 computed Active Time doesn't account for pod clock drift

itsmojo:


OmniKitUI/ViewControllers/OmnipodSettingsViewController.swift
```diff
@@ -459,10 +459,13 @@ class OmnipodSettingsViewController: RileyLinkSettingsViewController {
...
-                case .activatedAt:
+                case .activeTime:
                     let cell = tableView.dequeueReusableCell(withIdentifier: SettingsTableViewCell.className, for: indexPath)
-                    cell.textLabel?.text = LocalizedString("Active Time", comment: "The title of the cell showing the pod activated at time")
-                    cell.setDetailAge(podState.activatedAt?.timeIntervalSinceNow)
+                    cell.textLabel?.text = LocalizedString("Active Time", comment: "The title of the cell showing the pod active time")
+                    // Since the podState doesn't actually keep the pod's reported time active value returned in each response,
+                    // use a calculation based on a computed activated at time that is based on the dynamically updated expiresAt time
+                    // to have a consistent value and behavior when the pod's internal active time varies from the true elapsed time.
+                    cell.setDetailAge(podState.expiresAt?.addingTimeInterval(-Pod.nominalPodLife).timeIntervalSinceNow)
```
ps2, line 468: Hmm, I think for this field, which is time since pairing, we should probably show the actual time from the phone, which is real, synchronized, time. I think showing expiration time using the skewed pod clock makes a lot of sense using the updated fields, as you do want to know when it's going to expire, and that will give you the best idea.  But time since pairing should probably not be skewed based on pod drift.

itsmojo: The data reported in “Pod Settings” status section is displaying user friendly pod state (basically) as reported by the pod except for “Expires” which is a derived value that tracks the reported pod alive time and the “Active Time” which is derived even though it is actually reported by the pod (which is used to adjust the PodState.expiresAt value but not actually saved). Adjusting “Expires” using the pod’s reported active time makes it match the pod’s 72 hour expiration alert time, but the “Active Time” should also similarly match the pod’s values to prevent various anomalies with the “Pod Active” time not being an even hour value for the 72 hour expirationAdvsioryAlarm (in addition to being inconsistent with the “Expires” value), the 79 hour shutdownImminentAlarm, and the 80 hour end of service pod screamer. Additionally, if “Active Time” isn’t based the pod’s reported active time, it can differ from the Read Pod Status’s “Pod Active” value which is the pod’s reported active time (see https://github.com/ps2/rileylink_ios/issues/656 for an example). Another way to think about this is that a more natural implementation would have been to save the pod’s reported active time in PodState and just display this value (as a user might expect in this “Pod Settings” section), but this involves adding & maintaining another persistent PodState state value which I thought would be a harder sell since it wasn’t originally done this way instead of using a single one line change that has the same net result. Since all the other pod’s relative timing alarms & actions for the pod itself and in other places in Loop are all based on the pod’s notion of elapsed active time, there’s actually no good reason I can think of why the displayed “Active Time” should be based on the real, synchronized time instead of the pod’s reported active time. Additional note: this field isn’t actually the time since pairing, but rather the time since pod power up (i.e., insulin injection into the pod) which only the pod actually knows.

itsmojo:
Closed and replaced by https://github.com/ps2/rileylink_ios/pull/673 which includes other Active Time fixes (no refreshing and incorrect case name) and simply documents the current real Active Time calculation as well as an alternate calculation which would account for any pod clock drift. See conversation above for a discussion regarding this calculation.

PR #660, opened 2021-03-27
Bulletproof prime() to prevent 049 pod fault when incorrectly invoked

itsmojo:
Rework Pod Settings UI logic to prevent new pod flow with a faulted pod
Always invoke finalizeDoses() after the initial basal has been set
Rename PodState unfinishedPairing var to more appropriate unfinishedSetup
Remove unused and unneeded PodProgressState unfinishedPairing var

### DIFF
--- a/OmniKit/PumpManager/PodState.swift
+++ b/OmniKit/PumpManager/PodState.swift
@@ -115,7 +115,7 @@ public struct PodState: RawRepresentable, Equatable, CustomDebugStringConvertibl
         self.insulinType = insulinType
     }
     
-    public var unfinishedPairing: Bool {
+    public var unfinishedSetup: Bool {
         return setupProgress != .completed
     }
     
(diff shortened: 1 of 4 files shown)


OmniKit/PumpManager/PodState.swift
```diff
@@ -21,6 +21,10 @@ public enum SetupProgress: Int {
     case completed
     case activationTimeout
     case podIncompatible
+
+    public var primingNeverAttempted: Bool {
```
ps2, line 25: nit: maybe `primingNotStarted`; if we attempt a prime, but the pod doesn't hear it, this will still return true, even though we did attempt.

ps2 (approved):

OmniKit/PumpManager/PodState.swift
itsmojo: The `primingNeverAttempted` variable works as its name indicates and it is indeed the desired behavior. It will be false after setupProgress is set to `.startingPrime` immediately before attempting a prime bolus (independent of whether or not the pod actually hears it). Once priming is attempted, Loop should never send a FaultConfig command to this pod since the one time FaultConfig command has already been successfully completed and doing it again risks an 049 pod fault if the pod had heard the priming bolus attempt.

Issue #661, opened 2021-04-20
Continual "Pod already Primed" errors and other problems after quickly cancelling a Pair operation


itsmojo:
A user reports:
_________
Initially I made a user error when I clicked to prime my pod before I added insulin. I immediately realized and hit cancel realizing my mistake. (Not sure if this contributes to the error that occurred or just a fluke).

I added insulin to activate pod and proceeded by clicking prime.

As the pod was priming, my phone received the error message that I had lost connectivity and to move devices closer (everything was sitting next to each other though). Even with message, the pod kept priming. When it finished priming I canceled the “move pod closer” error message and it sent me back to this screen I captured when I tried to proceed.

The only option on this error screen was to prime (but when I clicked this it said my pod was already primed) or cancel and deactivate.

Without another option I swiped app off and immediately back on, and it let me proceed, recognized I had primed, and let me insert the cannula.
__

*** When asked if the app popped up a menu with a deactivate choice or if it went back to the previous view after the Cancel button was originally pressed, the user replied:

I don’t believe it originally took me back to the deactivate choice, but I’m not positive what screen it did take me too - I think it was just the initial pod filling screen.

_________

Here's the "Pod already Primed." screen at 4:08 local time:
[Sara_Goya_Pod_Pairing_Issue](https://user-images.githubusercontent.com/23020111/115180995-c28f4680-a08b-11eb-8b23-fe8762b151e8.jpg)

Here's the Loop Report:
[Loop-Report-2021-04-13-161605-0700.md](https://github.com/ps2/rileylink_ios/files/6333215/Loop-Report-2021-04-13-161605-0700.md)

Here's the relevant portion of the Omnipod messages with some formatting and commenting:

2021-04-13 23:06:52 noPod    send    ffffffff 0006 07041f050cb481d0 // AssignAddress no response (out of range?)
2021-04-13 23:07:02 noPod    send    ffffffff 0006 07041f050cb481d0
2021-04-13 23:07:10 noPod    receive ffffffff 0417 01150209000209000201000112120027adde9c1f050cb4037d // RSSI=$1c = 28 (too low for RSSI check)
2021-04-13 23:07:11 noPod    send    ffffffff 0006 07041f050cb481d0 // Loop retry on low RSSI value
2021-04-13 23:07:12 noPod    receive ffffffff 0417 01150209000209000202000112120027addea01f050cb48188 // RSSI=$20 = 21 (OK for Loop to proceed w/ pairing)
[...]

PR #662, opened 2021-04-20
Fix issue #661 "Pod already primed" and other errors after a cancel

itsmojo:
+ Change needsPairing() to return a simple Bool as warnings not handled
+ Rework needsPairing() to properly handle an already priming pod
+ Rename configureAndPrimeSession() to primeSession() to match reality
+ Remove unused OmnipodPumpManagerError enum types
+ Rework PodComms to prevent ever moving setupProgress backwards
+ Create new setupProgress.isPaired var & .podPaired for improved clarity

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -27,8 +27,6 @@ public protocol PodStateObserver: class {
 public enum OmnipodPumpManagerError: Error {
     case noPodPaired
     case podAlreadyPaired
-    case podAlreadyPrimed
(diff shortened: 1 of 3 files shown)


ps2:
Can you describe the user facing issue here? I.e. what problem does this solve?

itsmojo:
As detailed in https://github.com/ps2/rileylink_ios/issues/661, this PR solves the user problem where Loop gets into a state where Loop will continually print "Pod already Primed" each time the Prime button is pushed with no way to complete the pod startup operation (w/o restarting Loop) which can occur whenever a user quickly cancels a Prime operation.

PR #664, opened 2021-04-29
Move Pod Details to bottom, confirmationBeepsTapped() simplifications

itsmojo:


OmniKitUI/ViewControllers/OmnipodSettingsViewController.swift
```diff
@@ -712,27 +710,15 @@ class OmnipodSettingsViewController: RileyLinkSettingsViewController {
...
+                    let title = LocalizedString("Error setting confirmation beeps", comment: "The alert title for setting confirmation beeps error")
+                    self.present(UIAlertController(with: error, title: title), animated: true)
                 }
-                done()
-            })
-        } else {
-            pumpManager.setConfirmationBeeps(enabled: true, completion: { (error) in
-                if let error = error {
-                    DispatchQueue.main.async {
-                        let title = LocalizedString("Error enabling confirmation beeps", comment: "The alert title for enable confirmation beeps error")
```
ps2, line 729: Would prefer to keep the error message that the error happened while "enabling" or "disabling", vs just "setting".

ps2 (changes requested):
UI section ordering change is fine. I think the error message change is a step back, though.

OmniKitUI/ViewControllers/OmnipodSettingsViewController.swift
itsmojo: Agreed. Now using the original separate error strings for enabling and disabling confirmation beeps cases. This has the additional benefit of no additional internationalization work created by this PR.

PR #667, opened 2021-07-11
OrangeLink Patch 3.1 PR to Dev

jlucasvt:
PR with patch files for OrangeLink and OrangeLink Pro.  I received permission from BubbleDev and Vic Wu's team to push the PR to RLKit.   Pete - need some assistance on cleanup and possibly detecting when OL is present to disable the MySentry instead of commenting it out.

- Adds Battery Level Status Display (%)
- Adds OrangeLink Firmware (FW) and Hardware (HW) Version Listing
- Adds Battery Voltage (volts)
- Adds % Setting and Alert for Low Battery (Off or Set Point)
- Adds Voltage Setting and Alert for Low Voltage Alert (Off or Set Point)
- Adds Toggle to Enable/Disable Connection State 10 Second Blinking LED
- Adds Toggle to Enable/Disable Connection State Disconnect Vibration Alert
- Adds Test Switches to test Yellow and Red LED’s, and the Haptic Motor
- Adds [Find Device] Command (OrangeLink Pro Only)
- Fix for EMALink Communications Error when using Feature Patch
- Disables MySentry Packets to increase OrangeLink battery life with Medtronic x23/x54 pumps

Features

### DIFF
--- a/MinimedKit/Models/PumpModel.swift
+++ b/MinimedKit/Models/PumpModel.swift
@@ -51,7 +51,8 @@ public enum PumpModel: String {
     }
     
     public var hasMySentry: Bool {
-        return generation >= 23
+        //TODO: disable MySentry when using OL return generation >= 23
+        return false
     }
     
     var hasLowSuspend: Bool {
--- a/MinimedKitUI/RileyLinkMinimedDeviceTableViewController.swift
+++ b/MinimedKitUI/RileyLinkMinimedDeviceTableViewController.swift
@@ -53,6 +53,16 @@ public class RileyLinkMinimedDeviceTableViewController: UITableViewController {
         }
     }
     
+    private var fw_hw: String? {
+        didSet {
+            guard isViewLoaded else {
+                return
+            }
+            
+            cellForRow(.orl)?.detailTextLabel?.text = fw_hw
+        }
+    }
+    
     private var uptime: TimeInterval? {
         didSet {
             guard isViewLoaded else {
@@ -62,6 +72,17 @@ public class RileyLinkMinimedDeviceTableViewController: UITableViewController {
             cellForRow(.uptime)?.setDetailAge(uptime)
         }
     }
+    
+    private var battery: String? {
+        didSet {
+            guard isViewLoaded else {
+                return
+            }
+            
+            cellForRow(.battery)?.setDetailBatteryLevel(battery)
+        }
+    }
+
 
     private var lastIdle: Date? {
         didSet {
@@ -117,12 +138,69 @@ public class RileyLinkMinimedDeviceTableViewController: UITableViewController {
             } catch { }
         }
     }
+    
+    func updateBatteryLevel() {
+        device.runSession(withName: "Get battery level") { (session) in
+            let batteryLevel = self.device.getBatterylevel()
+            DispatchQueue.main.async {
+                self.battery = batteryLevel
+            }
+        }
+    }
+    
 
+    func orangeClose() {
+        device.runSession(withName: "Orange Action Close") { (session) in
+            self.device.orangeClose()
+        }
+    }
+    
+    func orangeReadSet() {
+        device.runSession(withName: "orange Read Set") { (session) in
+            self.device.orangeReadSet()
+        }
+    }
+    
+    func orangeReadVDC() {
+        device.runSession(withName: "orange Read Set") { (session) in
+            self.device.orangeReadVDC()
+        }
+    }
+
+    func writePSW() {
+        device.runSession(withName: "Orange Action PSW") { (session) in
+            self.device.orangeWritePwd()
+        }
+    }
+    
+    func orangeAction(index: Int) {
+        device.runSession(withName: "Orange Action \(index)") { (session) in
+            self.device.orangeAction(mode: index)
+        }
+    }
+    
+    func orangeAction(index: Int, open: Bool) {
+        device.runSession(withName: "Orange Set Action \(index)") { (session) in
+            self.device.orangeSetAction(index: index, open: open)
+        }
+    }
+    
+    func findDevices() {
+        device.runSession(withName: "Find Devices") { (session) in
+            self.device.findDevices()
+        }
+    }
+    
     private func updateDeviceStatus() {
         device.getStatus { (status) in
             DispatchQueue.main.async {
-                self.lastIdle = status.lastIdle
                 self.firmwareVersion = status.firmwareDescription
+                self.fw_hw = status.fw_hw
+                self.ledOn = status.ledOn
+                self.vibrationOn = status.vibrationOn
+                self.voltage = status.voltage
+                
+                self.tableView.reloadData()
             }
         }
     }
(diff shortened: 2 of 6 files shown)


MinimedKit/Models/PumpModel.swift
```diff
@@ -51,7 +51,8 @@ public enum PumpModel: String {
     }
     
     public var hasMySentry: Bool {
-        return generation >= 23
+        //TODO: disable MySentry when using OL return generation >= 23
```
ps2, line 54: This needs to be tackled at a higher level. PumpModels >= 23 should still return that they have MySentry capability, even when using OL, but the MinimedPumpManager could choose not to use it if using OL.

MinimedKitUI/RileyLinkMinimedDeviceTableViewController.swift
```diff
@@ -53,6 +53,16 @@ public class RileyLinkMinimedDeviceTableViewController: UITableViewController {
         }
     }
     
+    private var fw_hw: String? {
```
ps2, line 56: Naming conventions. `fw_hw` -> `firmwareHardware`

MinimedKitUI/RileyLinkMinimedDeviceTableViewController.swift
```diff
@@ -185,6 +267,27 @@ public class RileyLinkMinimedDeviceTableViewController: UITableViewController {
         updateRSSI()
         
         updateUptime()
+        
+        updateBatteryLevel()
```
ps2, line 271: Some of these should only be done when connected to OL.

ps2 (changes requested):
Thanks Jeremy. I haven't gone through the whole thing yet, but yes, it does need to be updated to be more conditional about whether or not the user is using an OL or a RL. If we can get that, I think we can land this; might have a few more changes after a full review.

ps2:
I'll take a look at updating this patch with the requested changes.

jlucasvt:
Thank you Pete.

MinimedKit/Models/PumpModel.swift
ps2: There is a "Use MySentry" option in pump settings in dev that can be used to enable/disable use of MySentry. The battery tradeoffs are similar on both RL and OL.

RileyLinkBLEKit/RileyLinkDevice.swift
```diff
@@ -338,18 +415,86 @@ extension RileyLinkDevice: PeripheralManagerDelegate {
...
+                          UserDefaults.standard.double(forKey: "voltage_alert_value") != 0 else { return }
+                    
+                    UserDefaults.standard.setValue(Date().timeIntervalSince1970, forKey: "voltage_date")
+                    DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
+                        let value = UserDefaults.standard.double(forKey: "voltage_alert_value")
+                        if (Double(self.voltage) ?? 100) <= value {
+                            let content = UNMutableNotificationContent()
+                            content.title = "Low Voltage"
+                            content.subtitle = self.voltage
+                            let request = UNNotificationRequest.init(identifier: "Orange Low Voltage", content: content, trigger: nil)
```
ps2, line 469: Should not be triggering UNNotificationRequests from device code

RileyLinkBLEKit/RileyLinkDevice.swift
```diff
@@ -338,18 +415,86 @@ extension RileyLinkDevice: PeripheralManagerDelegate {
...
+            manager.queue.async {
+                UserDefaults.standard.setValue(Date().timeIntervalSince1970, forKey: "battery_date")
+                let batteryLevel = self.getBatterylevel()
+                DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
+                    let value = UserDefaults.standard.integer(forKey: "battery_alert_value")
+                    if (Int(batteryLevel) ?? 100) <= value {
+                        let content = UNMutableNotificationContent()
+                        content.title = "Low Battery"
+                        content.subtitle = batteryLevel
+                        let request = UNNotificationRequest.init(identifier: "Orange Low Battery", content: content, trigger: nil)
```
ps2, line 492: Should not be triggering UNNotificationRequests from device code

bartsopers:
@ps2 do you have an idea for how to detect that an OL is present? If so, I'd be interested in your solution so we can integrate that in AAPS too.

ps2:
> @ps2 do you have an idea for how to detect that an OL is present? If so, I'd be interested in your solution so we can integrate that in AAPS too.

With newer OL firmware, there are OL specific attributes present. For older OL firmware, I think that detecting the secure dfu attribute will be sufficient to distinguish an OL from a RL.

RileyLinkBLEKit/RileyLinkDevice.swift
```diff
@@ -338,18 +415,86 @@ extension RileyLinkDevice: PeripheralManagerDelegate {
...
+                    
+                    guard Date() > Date(timeIntervalSince1970: UserDefaults.standard.double(forKey: "voltage_date")).addingTimeInterval(60 * 60),
+                          UserDefaults.standard.double(forKey: "voltage_alert_value") != 0 else { return }
+                    
+                    UserDefaults.standard.setValue(Date().timeIntervalSince1970, forKey: "voltage_date")
+                    DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
+                        let value = UserDefaults.standard.double(forKey: "voltage_alert_value")
+                        if (Double(self.voltage) ?? 100) <= value {
+                            let content = UNMutableNotificationContent()
+                            content.title = "Low Voltage"
```
ps2, line 467: Strings should be localized.

RileyLinkBLEKit/RileyLinkDevice.swift
```diff
@@ -338,18 +415,86 @@ extension RileyLinkDevice: PeripheralManagerDelegate {
...
+    }
+    
+    func resetBatteryAlert() {
+        guard Date() > Date(timeIntervalSince1970: UserDefaults.standard.double(forKey: "battery_date")).addingTimeInterval(60 * 60) else { return }
+        if UserDefaults.standard.integer(forKey: "battery_alert_value") != 0 {
+            manager.queue.async {
+                UserDefaults.standard.setValue(Date().timeIntervalSince1970, forKey: "battery_date")
+                let batteryLevel = self.getBatterylevel()
+                DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
+                    let value = UserDefaults.standard.integer(forKey: "battery_alert_value")
```
ps2, line 487: Alert settings should be stored in PumpManager state.

RileyLinkBLEKit/PeripheralManager+RileyLink.swift
```diff
@@ -295,7 +345,161 @@ extension PeripheralManager {
...
+            } catch (_) {
+                add(log: "orangeAction failed")
+            }
+        }
+        if mode == .off, mode == .shakeOff {
+            orangeClose()
+        }
+    }
+    
+    func findDevices() {
```
ps2, line 404: Some of these functions seem to not be supported in early versions of OL firmware. We should hide them if they don't do anything, and people are still using those versions.

ps2:
Ongoing work to integrate support into dev is going on here: https://github.com/ps2/rileylink_ios/pull/671

RileyLinkBLEKit/RileyLinkDevice.swift
```diff
@@ -338,18 +415,86 @@ extension RileyLinkDevice: PeripheralManagerDelegate {
                     self.log.error("Skipping parsing characteristic value update due to missing BLE firmware version")
                 }
 
+                self.resetBatteryAlert()
+                self.orangeReadVDC()
                 self.assertIdleListening(forceRestart: true)
             }
         case .responseCount?:
             // PeripheralManager.Configuration.valueUpdateMacros is responsible for handling this response.
+            self.resetBatteryAlert()
```
ps2, line 424: Every single packet is triggering multiple battery level checks.

PR #669, opened 2021-07-12
Fixes to better handle comms optimizations with unreliable comms

itsmojo:
+ Add non-persistent var to denote when to skip comms optimizations
+ Set skipNextCommsOptimization on init or for any error or unexpected
  message response to better handle bad comms on restarts and out of range
+ Have updateDeliveryStatus() create an unfinalizedBolus if pod is bolusing
  and PodState doesn't have one to handle both types of status responses
+ Have updateDeliveryStatus() set skipNextCommsOptimization when the pod
  is bolusing or TB is active & it doesn't match the PodState's status
+ Some bolus communication optimization logic improvements
+ Add patchable variable to disable communication optimizations


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -13,6 +13,7 @@ import RileyLinkBLEKit
 import UserNotifications
 import os.log
 
+fileprivate let disableCommsOptimizations = false
```
ps2: Please do not add static debug or config options to the code.

ps2 (changes requested):
I'm not sure about accepting the optimization here; having trouble following the logic. The config var needs to go, though. If the optimization should be in the code, it should be in. If not, it should be out. Adding a flag to the code makes the code harder to reason about.

Capturing an active bolus not registered in PodState looks ok, and I can merge that in if you could submit that separately.

itsmojo:
Re "I'm not sure about accepting the optimization here; having trouble following the logic". This PR is not an optimization, but an important fix to better handle a number of potential edge cases by avoiding doing the previously merged bolus comms optimization in a number of additional cases that can occur with unreliable communications. Please let me know what logic you are still having trouble following after reading the various comments and logging for the various cases.

itsmojo:
Closed and replaced by https://github.com/ps2/rileylink_ios/pull/672 to deal with github merge issues. Grrr...

PR #672, opened 2021-07-26
Fixes to better handle Omnipod comms optimizations with unreliable comms

itsmojo:
+ Add non-persistent var to denote when to skip comms optimizations
+ Set skipNextCommsOptimization on init or for any error or unexpected
message response to better handle bad comms on restarts and out of range
+ Have updateDeliveryStatus() create an unfinalizedBolus if pod is bolusing
and PodState doesn't have one to handle both types of status responses
+ Have updateDeliveryStatus() set skipNextCommsOptimization when the pod
is bolusing or TB is active & it doesn't match the PodState's status
+ Skip bolus communication optimizations for manual boluses

### DIFF
--- a/OmniKit/PumpManager/PodState.swift
+++ b/OmniKit/PumpManager/PodState.swift
@@ -104,6 +104,8 @@ public struct PodState: RawRepresentable, Equatable, CustomDebugStringConvertibl
     
     public var insulinType: InsulinType
     
+    public var skipNextCommsOptimization: Bool
+
     public init(address: UInt32, piVersion: String, pmVersion: String, lot: UInt32, tid: UInt32, packetNumber: Int = 0, messageNumber: Int = 0, insulinType: InsulinType) {
         self.address = address
         self.nonceState = NonceState(lot: lot, tid: tid)
@@ -121,6 +123,7 @@ public struct PodState: RawRepresentable, Equatable, CustomDebugStringConvertibl
         self.setupProgress = .addressAssigned
         self.configuredAlerts = [.slot7: .waitingForPairingReminder]
         self.insulinType = insulinType
+        self.skipNextCommsOptimization = true // skip any initial comms optimizations
     }
     
     public var unfinishedSetup: Bool {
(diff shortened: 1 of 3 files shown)


OmniKit/PumpManager/PodState.swift
```diff
@@ -104,6 +104,8 @@ public struct PodState: RawRepresentable, Equatable, CustomDebugStringConvertibl
     
     public var insulinType: InsulinType
     
+    public var skipNextCommsOptimization: Bool
```
ps2: Having control-flow variables at this wide scope strikes me as a bad pattern. It creates a situation where the code branches based on whether it is set or not, but the name offers no clue as to *why* the branch is going one way or the other.

State variables should convey something about the state of the program, and describe the state, rather than describing how code should flow. The code can branch in reaction to the state.

Is there some way you can change this to describe the actual state you're wanting to modify behavior for?  I.e. if you want to do something special when the last request errored, it'd be much better to store a variable called `lastRequestErrored`, than a flag called `doSomethingSpecial`.  And then the code can read

```
if lastRequestDidError {
    doSomethingSpecial()
}
```

ps2 (changes requested):
I think I understand now that you're trying to force a getStatus before bolusing in some states, but I'm not sure I understand those states completely. I think that having a "control-flow" variable at PodState scope is making this tough to reason about.

OmniKit/PumpManager/PodState.swift
itsmojo: The original prototype this variable was actually named `lastCommsFailed`. But this variable's use has evolved and it is now set for the following three distinct cases and cleared only upon Loop receiving an expected response:

+ Whenever Loop doesn't receive the expected response for any message it sends
+ Whenever the pod's bolus or temp basal delivery status doesn't match Loop's
+ On Loop app start up

For all of these cases, any pod comms optimizations (i.e., skipping the get status command before a bolus and later skipping the cancel TB command when going from the scheduled basal to a TB) should not be performed for the best/natural behavior in the face of unreliable comms. Thus I thought that using `skipNextCommsOptimization` was a good way to succinctly describe this variable's use (and upon reflection it actually describe the state of the program). If you still don't like the variable name which I decided on after some thought (and with no consultation which I had requested), please suggest a suitable alternate variable name which would encapsulate its various uses/state as I'd like to get this PR to land ASAP. Thanks!

Would a comment on the variable declaration documenting the various places it is used be helpful?

PR #680, opened 2021-08-31
Remove extra period from Device Error message

alex-mchardy:
Removed period from end of string. Two periods are shown on error without this removed. 

IMG_8386


ps2:
Thanks!

PR #685, opened 2021-09-29
Add new Automatic Bolus Beeps button

itsmojo:


ps2 (commented):
Would like to revisit this when/if we get the new ui backported to Eros pods.

Issue #682, opened 2021-09-20
Add v2.2.5 improvements for RL to dev branch


marionbarker:
EmaLink could not connect without power cycle of phone and EmaLink.
This may be similar to an issue fixed with v2.2.5 version of the code.
Zulipchat report:
* https://loop.zulipchat.com/#narrow/stream/144182-development/topic/dev.3A.20Ema.20question/near/253983158

Copy of that zulipchat message for completeness:

I captured an Xcode log for a case where both real EmaLink and a RileyLink running EmaLink firmware had a problem with dev code. Stopped being able to communicate with pump with either link using my "real" Loop dev phone.

Tried without success:
* power cycle link
* close and restart Loop

Specific devices having problems (both devices show subg and ble on backup phone running v2.2.6):
* A: RileyLink hardware with EmaLink firmware (subg_rfspy 2.2.19 / ble_rfspy 3.11) - showing only ble with dev
* B: EmaLink (subg_rfspy 2.2 / ble_rfspy 3.17) - showing only ble with dev

Steps:
* I built Loop dev (no patches for rileylink_ios) onto "real" phone at 2021-09-19 16:59 so I could capture the log.
* Both A and B were showing the ble_rfspy only before I built Loop dev on top of existing Loop dev app.
* I touched A in pod settings menu: and it then reported both subg as well as ble firmware. (still 16:59)
* I touched B in pod settings menu: and it remained unable to communicate. (17:00)
* Tried power cycle of B and still only ble showed.
* Disconnected from Xcode.  Power cycled phone.  Had to also power cycle B before I could connect.  

[20210919_1659_to_1703_dev_Xcode_debug_log_EL.txt](https://github.com/ps2/rileylink_ios/files/7193197/20210919_1659_to_1703_dev_Xcode_debug_log_EL.txt)

marionbarker:
I looked at the modifications that were implemented to fix the "ema" problem and to clean up some "orange" comms leading up to loop-release/v2.2.5 (commit 15d1997 of rileylink_ios).

I believe the attached patch captures the modifications needed for dev branch to have these same improvements.

I have been using this patch successfully on dev branch on myself using EmaLink and OrangeLink Pro for more than one week. The naming of the patch might need some explanation.  The "master" referred to in the name is the LoopWorkspace build for master that points to rileylink_ios commit 15d1997.

[...]

PR #688, opened 2021-10-04
Issue 682: add v2.2.5 improvements for RL to dev

marionbarker:
Apply the patch proposed to solve Issue 682
* https://github.com/ps2/rileylink_ios/issues/682

### DIFF
--- a/RileyLinkBLEKit/Response.swift
+++ b/RileyLinkBLEKit/Response.swift
@@ -103,7 +103,7 @@ struct GetVersionResponse: Response {
     let version: String
 
     init?(data: Data) {
-        guard data.count > 1, let code = ResponseCode(rawValue: data[data.startIndex]) else {
+        guard data.count > 0, let code = ResponseCode(rawValue: data[data.startIndex]) else {
             return nil
         }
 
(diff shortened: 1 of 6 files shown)


elnjensen:
Just to add a data point in support of this PR, I was unable to get my EmaLink to work stably with dev before I applied these changes. Since then, it has worked well.  I don’t know the technical details of what is changed here, but empirically these changes are working for me.

ps2:
This is mostly good, as it's taking updates I applied to master and copying them to dev, but it is backing out some dev changes that are good (crash fixes) that are not in master. I'll try to get a version of this applied to dev soon.

RileyLinkBLEKit/Response.swift
```diff
@@ -103,7 +103,7 @@ struct GetVersionResponse: Response {
     let version: String
 
     init?(data: Data) {
-        guard data.count > 1, let code = ResponseCode(rawValue: data[data.startIndex]) else {
+        guard data.count > 0, let code = ResponseCode(rawValue: data[data.startIndex]) else {
```
ps2, line 106: This, and the following similar changes, are re-introducing crashes.

ps2 (changes requested):

marionbarker:
@ps2 I think you are letting me know that you will make the required changes and that you don't need action from me. Thank you for fixing this.

PR #689, opened 2021-10-30
Improved PodCommsError fault checking during pod setup

itsmojo:
Fix compiler error when uncommenting mock cannula insertion fault code
Rework mock cannula insertion fault code to not be within a comment
Change mock pod fault during pairing to be a more realistic fault type
Remove extra "." in OmnipodPumpManagerError.notReadyForCannulaInsertion description
Default insulin type to .novolog in all places to be consistent

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -38,7 +38,7 @@ extension OmnipodPumpManagerError: LocalizedError {
         case .podAlreadyPaired:
             return LocalizedString("Pod already paired", comment: "Error message shown when user cannot pair because pod is already paired")
         case .notReadyForCannulaInsertion:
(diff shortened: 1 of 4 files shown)


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -647,12 +647,14 @@ extension OmnipodPumpManager {
         
         #if targetEnvironment(simulator)
         let mockDelay = TimeInterval(seconds: 3)
+        let mockFaultDuringInsertCannula = false
```
ps2, line 650: Please avoid "configuration in code". If it's truly something that needs to be user configurable, it should go into the UI. Otherwise, make the decision and implement it as simply as possible.

itsmojo: This was simply a way to avoid having commented out optional developer test code which can drift out of date. This isn't something for users to configure and it doesn't make sense to make the decision to always fault when using the simulator.

Issue #686, opened 2021-10-01
Updates Needed for Ema and Orange Displays


marionbarker:
This comment was updated (Oct 12, 2021) to include vs2 of the patch and with easier usage instructions (Oct 16, 2021).

This comment was updated on July 5, 2022 with the following changes:
* This code was added to dev but not master
  * Loop 2.2.9 needs the patch to see EmaLink `Battery level` or Orange Link Pro `Find Device` in the RileyLink display.
* The rileylink_ios commit level for a fresh download is now 1651753 - and patch still works. 
  * I did not rename the patch, but have updated the instructions where it says, must be at 15d1997, to now say: must be at 15d1997 or 1651753.
* I have removed references to running this on dev because no patch is needed with dev.

## Purpose of Patch:

The items displayed in the RileyLink Display need some updates to handle:
* OrangeLink and OrangeLink Pro
  * Interpret hardware type for all released versions of OL and OL Pro firmware
  * Show the "Find Device" line for HW1.1 (OL Pro) regardless of firmware
* EmaLink
  * Method to identify EmaLink
  * Battery display
  * Add battery alert level picker
  * Change label to "Invert LED Logic" (actual function of "Diagnostic LEDs" row for .riley)

The battery alert picker sets a common level for all .ema and .orange
* If you switch between these two, you may choose to adjust the level alert for whichever device is currently selected

## Step 0: Open Terminal at LoopWorkspace Folder

For each command example shown below, there is a copy button that appears when you hover your mouse to the upper right corner of the example.  You can tap on the copy button and then paste it into your terminal.

You must in be the LoopWorkspace folder to use these commands. Each set of commands, in later steps, returns you to LoopWorkspace.  (If there is an error, you probably need to start over.)

If you are running dev (see notes below)

Open a terminal window with path at the LoopWorkspace folder used for the build. 

You can use Finder to do this (these instructions assume you used the Build-Select Script - if your files are in a different folder, make the appropriate adjustment)
1. Use Finder to navigate to Downloads / BuildLoop and open the folder with the most recent date (e.g., Loop-Master-211006-0524)
1. Right click (or hold down control key and click) on LoopWorkspace folder
1. In drop-down menu, select New Terminal at Folder

This opens a new terminal in the correct folder.  You will use this terminal window for Steps 1 - 3.

But first, make sure you are in the correct directory, copy and paste this command into the terminal.

```
pwd
```

[...]

PR #690, opened 2021-11-07
ema and orange update, see issue 686

marionbarker:
This PR will prevent users from having to apply the patch from issue #686 for Loop v2.2.6 master to get these features:

OrangeLink Pro
* Find Device 

EmaLink
* Battery Display
* Battery Notification


ps2:
Not currently adding new features to release branch, but this is small enough, that if we have another urgent fix for release, I'll try to get it added. Thanks!

PR #691, opened 2021-11-21
Bolus auto resume fixes and updates

itsmojo:
* Cancel any pod suspend alerts for auto resume on bolus
* Fix string formatting for debug error logging
* Never auto resume for an automatic bolus
* Add autoResumeOnManualBolus let variable for user control
* Add PodCommsError recoverySuggestion for podSuspended

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -1378,18 +1378,28 @@ extension OmnipodPumpManager: PumpManager {
                 state.bolusEngageState = .engaging
             })
 
-            // If pod suspended, resume basal before bolusing to match existing Medtronic PumpManager behavior
+            // Initialize to true to match existing Medtronic PumpManager behavior for any
+            // manual boluses or to false to never auto resume a suspended pod for any bolus.
+            let autoResumeOnManualBolus = true
+
             if case .some(.suspended) = self.state.podState?.suspendState {
+                // Pod suspended, only auto resume for a manual bolus if autoResumeOnManualBolus is true
+                if automatic || autoResumeOnManualBolus == false {
+                    self.log.error("enactBolus: returning pod suspended error for %@ bolus", automatic ? "automatic" : "manual")
+                    completion(.deviceState(PodCommsError.podSuspended))
+                    return
+                }
                 do {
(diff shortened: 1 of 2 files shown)


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1378,18 +1378,28 @@ extension OmnipodPumpManager: PumpManager {
                 state.bolusEngageState = .engaging
             })
 
-            // If pod suspended, resume basal before bolusing to match existing Medtronic PumpManager behavior
+            // Initialize to true to match existing Medtronic PumpManager behavior for any
+            // manual boluses or to false to never auto resume a suspended pod for any bolus.
+            let autoResumeOnManualBolus = true
```
ps2, line 1383: nit: please don't add static configuration variables. makes line 1387 more complicated to parse, and if this is a decision we want to make in the code, it should be made authoritatively.

ps2 (approved):
I think this idea is ok, since it matches mdt. Just a comment about avoiding unneeded complicating variables.

marionbarker:
I was not aware of this feature until Joe sent a PR into loopnlearn. After testing, I really like the auto resume when doing a manual bolus from a suspended pump. IMO that should be the default because you are probably doing a bolus because BG is going up and auto resume with manual bolus saves the user a step. If that’s the decision then I’ll add documentation for this in LoopDocs.

I think auto resume on manual bolus will work nicely with alerts after pump has been suspended for a user selected amount of time.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: Most parents I have discussed the auto resume on bolus feature with would prefer to not have this potentially risky behavior. But since I was told to maintain this behavior for Omnipod, I wanted to document this non-obvious behavior in a way which allows for its convenient disabling.

itsmojo:
N.B., This PR doesn’t actually change the long standing auto resume on bolus behavior other than to expressly limit it to manual boluses and to cancel any suspend alerts when doing the resume. In the past, the Loop automatic bolus logic would do an automatic bolus on a suspended pod which resulted in frustration and panic from users handling a severe hypoglycemia incident. Limiting the auto resume to just manual boluses provides additional protection in case this type of automatic bolus logic error ever happens again.

PR #692, opened 2021-11-21
Add option for 95% battery alert

Nnk2:
Adding 95% battery alert option, for timely alert for (disposable lithium) batteries that stay at ~100% until empty. 

NB: Please check this PR, since this is my first PR ever.


ps2:
What kind of RL are you using? If you have a RL designed for different batteries, it should maybe be reporting different %s

ps2:
Please see above. I would like to understand more about this use case. Reopen when there is more info on your use case.

PR #698, opened 2022-01-31
Avoid unneeded Omnipod command to set a TBR from scheduled basal

itsmojo:


ps2:
Is this actually starting a new temp basal on top of an existing temp basal w/o canceling the previous one? Does that work? 

And then the comments and title keep confusing me. The "optimization" here is skipping doing something, "cancel" means to stop doing something. And the title/comments are written to indicate we're skipping that skipping. So it reads like double negatives about a negating operation.  "Skipping the cancel TB skipping"??? Any way to clean up these comments to avoid the mental gymnastics? Rewrite in a positive sense?

itsmojo:
* Is this actually starting a new temp basal on top of an existing temp basal w/o canceling the previous one? Does that work?

Obviously starting a new temp basal on top of an existing temp basal will not work as the existing temp basal must first be cancelled or the pod will fault. The original title says "from scheduled basal" and the commenting says to "Do the cancel TB command if ... we currently have a temp basal running", so I'm not sure why this is even being asked here.

+ "cancel" means to stop doing something. And the title/comments are written to indicate we're skipping that skipping. So it reads like double negatives about a negating operation. "Skipping the cancel TB skipping"??? 

Doing a cancel TB command is an action to be performed, not a "skipping". In the right circumstances with this PR, we are avoiding an action to send an unneeded command (a cancel TB command when there is no possible active TB to cancel).

+ Any way to clean up these comments to avoid the mental gymnastics? Rewrite in a positive sense?

Will attempt to do this later today after I'm back home.

ps2:
> On Jan 31, 2022, at 9:44 AM, Joe Moran ***@***.***> wrote:
> 
> 
> Is this actually starting a new temp basal on top of an existing temp basal w/o canceling the previous one? Does that work?
> Obviously starting a new temp basal on top of an existing temp basal will not work as the existing temp basal must first be cancelled or the pod will fault. The original title says "from scheduled basal" and the commenting says to "Do the cancel TB command if ... we currently have a temp basal running", so I'm not sure why this is even being asked here.
> 
I guess I misread the code. Being honest here. This might indicate the logic is getting too complicated.
> "cancel" means to stop doing something. And the title/comments are written to indicate we're skipping that skipping. So it reads like double negatives about a negating operation. "Skipping the cancel TB skipping"???
> Doing a cancel TB command is an action to be performed, not a "skipping". In the right circumstances with this PR, we are avoiding an action to send an unneeded command (a cancel TB command when there is no possible active TB to cancel)
> 
The two “skippings” are 1) the actual “skipping” in the comment and the "avoid always” in the title and 2) the optimization, which is a skipping of some part of the normal procedure being done.  The “cancel” is another term that makes you have to think about the object of the verb, the TB, as being undone. So, just a lot of negative wording. In each of those three things, you need to think of 1) the object, and 2) negate its sense of meaning.
> Any way to clean up these comments to avoid the mental gymnastics? Rewrite in a positive sense?
> Will attempt to do this later today after I'm back home.
>

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1580,30 +1580,46 @@ extension OmnipodPumpManager: PumpManager {
...
-            case .uncertainFailure(let error):
-                // TODO: Return PumpManagerError.uncertainDelivery and implement recovery
-                completion(.communication(error))
-                return
-            case .success(let cancelTempStatus, _):
-                status = cancelTempStatus
-            }
+            // Skip the Cancel TB comms optimization if the last message had any
+            // comms issues or if the last delivery status hasn't been verified OK
+            let skipCancelTBCommsOptimization = self.state.podState?.lastCommsOK == false ||
```
ps2: Maybe rename `skipCancelTBCommsOptimization` to `needStatusCheck`?

OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -1580,30 +1580,46 @@ extension OmnipodPumpManager: PumpManager {
...
+                self.state.podState?.deliveryStatusVerified == false
 
-            guard !status.deliveryStatus.bolusing else {
-                completion(.communication(PodCommsError.unfinalizedBolus))
-                return
-            }
+            // Do the cancel TB command if we are resuming a normal basal,
```
ps2: "a normal basal" -> "scheduled basal delivery"?

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: I changed this variable to uncertainDeliveryStatus as it more appropriate. We don't need to a status check in these cases, we just need to do the cancel temp basal command because we can't be certain of the current delivery status. The skipCancelTBCommsOptimization variable name was originally chosen to try make it easier for someone who wanted to disable this optimization.

OmniKit/PumpManager/OmnipodPumpManager.swift
itsmojo: Done. Also changed the name of local variable from resumingNormalBasal to resumingScheduledBasal.

ps2:
Thanks, I think that reads more clearly.

PR #707, opened 2022-03-10
DetailedStatus updates for better consistency with OmniBLE

itsmojo:
+ previousPodProgressStatus now comes from the VV byte
+ YYYY word returned as new possibleFaultCallingAddress field
+ limit debugDescription string to only potentially useful values


ps2:
Thanks!

PR #709, opened 2022-03-13
enable Eros pod lifecycle display

marionbarker:
Joe noticed there was no lifecycle display for an Eros pod that had been running 75 hours.
This change enables the display.
He also noted there had been no HUD indication of expiration.  I wasn't sure how to fix that one.

Note - no modification was made to when display begins, i.e., display starts at 24 hours before expiration.
Changes as discussed in [zulipchat](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Easy.20access.20to.20Pod.20Age/near/275112771)(for Eros and Dash) will be a separate PR.


ps2:
Thanks!

PR #711, opened 2022-03-15
Gracefully handle boluses below minimum increment

elnjensen:
Display error message rather than assertionFailure when manual bolus entered is below minimum deliverable increment.  Addresses issue LoopKit/Loop#1637.


ps2:
Calling enactBolus with 0 units is a programming error, so we want to catch this in development.

PR #723, opened 2022-06-03
Various Omnipod Eros Confirmation Beep fixes and improvements

itsmojo:
+ Automatic Bolus Beeps renamed to Extended Confirmation Beeps for better consistency with Dash implementation
+ Update Extended Confirmation Beeps to beep for any automatic delivery changes made by Loop
(i.e., for set & cancel temp basal in addition to automatic boluses) for better consistency and usefulness
+ Include basic manual temp basal beep support for if/when this feature is ever added for Omnipod Eros in Loop
+ Fix commands using a beep block causing an incorrect completion beep for an in-progress automatic bolus
+ Fix enabling Confirmation Beeps causing an incorrect completion beep for an in-progress automatic bolus
+ Remove use of basal schedule completion beeps which are meaningless since they never normally complete

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -390,6 +390,48 @@ extension OmnipodPumpManager {
         return .noBolus
     }
     
+    // Returns true if there an unfinishedDose for a manual bolus (independent of whether it is finished)
+    private var hasUnfinalizedManualBolus: Bool {
+        if let automatic = state.podState?.unfinalizedBolus?.automatic, !automatic {
+            return true
+        }
+        return false
+    }
+
+    // Returns true if there an unfinishedDose for a manual temp basal (independent of whether it is finished)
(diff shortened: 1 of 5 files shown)


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -390,6 +390,48 @@ extension OmnipodPumpManager {
...
+        return beepMessageBlock
+    }
+
+    private var shouldBeepForAutomaticBolus: Bool {
+        return self.confirmationBeeps && self.extendedBeeps
+    }
+
+    private var shouldBeepForAutomaticTempBasal: Bool {
+        let automaticTempBasalBeeps = true
+        return self.confirmationBeeps && self.extendedBeeps && automaticTempBasalBeeps
```
ps2: Same comment here as in Dash PR. Could you simplify this to remove the no-op var?

itsmojo: Done

PR #733, opened 2022-06-28
ReservoirLevel and various MesageBlocks improvements

itsmojo:
+ Update ReserviorLevel to not depend on matching an exact FP value
+ Update MessageBlock pulse to units calcs for more accurate FP value
+ Update DetailedStatus to match OmniBLE
+ Update VersionResponse to match OmniBLE
+ Use ProductId instead of ProductID for better consistency


OmniKit/MessageTransport/MessageBlocks/VersionResponse.swift
```diff
@@ -11,6 +11,9 @@ import Foundation
 fileprivate let assignAddressVersionLength: UInt8 = 0x15
 fileprivate let setupPodVersionLength: UInt8 = 0x1B
 
+let erosProductId: UInt8 = 2        // 2nd gen Eros Omnipod (PM=PI=2.x.y)
```
ps2: I don't see these used anywhere?

ps2 (commented):
Looks good. Just some unused vars to clean up?

OmniKit/MessageTransport/MessageBlocks/VersionResponse.swift
itsmojo: Right now the OmniKitTests should be using this value for its VersionResponse tests. I was planning on doing this at a later time, but could do this now if desired (either as part of this PR or as a separate one -- I'd prefer the later).

OmniBLE already has the productId in its PodState while Eros does not (we didn't understand what this byte represented back then). I was thinking that in the future the productId could be used to tell what the Omnipod type is when we are actually sharing code under OmnipodCommon across different Omnipod types.

PR #740, opened 2022-08-11
Changes to better match OmniBLE and updated VersionResponse

itsmojo:
+ Update ErrorResponse comments to match OmniBLE
+ Rename VersionResponse pmVersion to firmwareVersion to match OmniBLE
+ Rename VersionResponse prVersion to iFirmwareVersion to match OmniBLE
+ Add missing check for out of range reservoir updates to match OmniBLE
+ Update PodState firmware ordering to match VersionResponse layout
+ Improved and cleaned up PodStateTests using more VersionResponse fields

### DIFF
--- a/OmniKit/PumpManager/OmnipodPumpManager.swift
+++ b/OmniKit/PumpManager/OmnipodPumpManager.swift
@@ -169,7 +169,10 @@ public class OmnipodPumpManager: RileyLinkPumpManager {
             }
 
             if oldValue.podState?.lastInsulinMeasurements?.reservoirLevel != newValue.podState?.lastInsulinMeasurements?.reservoirLevel {
-                if let lastInsulinMeasurements = newValue.podState?.lastInsulinMeasurements, let reservoirLevel = lastInsulinMeasurements.reservoirLevel {
+                if let lastInsulinMeasurements = newValue.podState?.lastInsulinMeasurements,
+                   let reservoirLevel = lastInsulinMeasurements.reservoirLevel,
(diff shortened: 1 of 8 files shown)


OmniKit/PumpManager/OmnipodPumpManager.swift
```diff
@@ -169,7 +169,10 @@ public class OmnipodPumpManager: RileyLinkPumpManager {
             }
 
             if oldValue.podState?.lastInsulinMeasurements?.reservoirLevel != newValue.podState?.lastInsulinMeasurements?.reservoirLevel {
-                if let lastInsulinMeasurements = newValue.podState?.lastInsulinMeasurements, let reservoirLevel = lastInsulinMeasurements.reservoirLevel {
+                if let lastInsulinMeasurements = newValue.podState?.lastInsulinMeasurements,
+                   let reservoirLevel = lastInsulinMeasurements.reservoirLevel,
+                   reservoirLevel != Pod.reservoirLevelAboveThresholdMagicNumber
```
ps2, line 174: It's a good change here to not report reservoir to Loop when we don't have a reservoir reading. I do think it might be better to keep `StatusResponse.reservoirLevel` as the magic number (rather than nil as was a while back), but to not propagate that beyond that structure.  `PodInsulinMeasurements.reservoirLevel` is a `Double?`, and users of that struct should either 1) know what the magic number is and how to interpret it, or 2) be able to handle missing values.  I don't think they should have to know 1 *and* 2.

ps2 (approved):
Thanks! I did notice some redundant type space, and have a comment for how we could simplify, but not needed to get this landed. Will land this now.

PR #745, opened 2022-09-15
Fixes and improvements to MinimedPumpManager

ps2:
* Improved logging of comms
* Avoid multiple retries of temp basal issuance that led to [multiple back-to-back temp basals being issued](https://github.com/LoopKit/Loop/issues/1790)
* Add short sleep between message sends to avoid getting repeated response from last command (due to this issue: https://github.com/LoopKit/Loop/issues/1791)
* Fix issue where 0-temp could be mistaken for temp basal cancellation during reconciliation
* Short pump event names, for insulin delivery view
* Fix localization issues.
* Loop issued events were being finalized without reconciliation could result in non-loop-issued pump events being skipped; i.e. miss fetching events that were in the pump history.


ps2:
Fixes https://github.com/LoopKit/Loop/issues/1789 and https://github.com/LoopKit/Loop/issues/1790, and reduces problems with https://github.com/LoopKit/Loop/issues/1791

PR #747, opened 2022-09-24
MDT Pump Settings command improvements

itsmojo:
+ Only display "MySentry Pair" command for mySentry capable pumps
+ Move "MySentry Pair" command to list end as per "Use MySentry"
+ Fix broken "Read Pump Status" battery voltage formatting
(e.g., display "1.54 volts" instead of "1.5400 V volts")


ps2:
The debugging commands are going to be removed from this interface, as they are on master.

PR #752, opened 2022-11-27
For HUD lifecycle display trigger, use scheduledExpirationReminderOffset

marionbarker:
Modify the trigger for when lifecycle is displayed in HUD taking into account state.scheduledExpirationReminderOffset.
Matches the change in [OmniBLE PR 40](https://github.com/LoopKit/OmniBLE/pull/40).
The OmniBLE modification was tested with rPI DASH simulator.
This modification builds but was not tested.


marionbarker:
Modified using the pattern tested with DASH so that Pod lifecycle color changes to guidanceColors.warning at the same time lifecyle bar shows up in the HUD.

This was not tested on Eros pods - code change done by inspection.

PR #759, opened 2023-01-07
Remove Crypto target

bill-foreflight:
When integrating Caregiver, I was running into collisions due to the Crypto target name matching an Apple Crypto library. This code is used for forming a sha1 using the CommonCrypto library. 

Rather than renaming the library, I thought it may be best to remove the target and just put the 19 lines of needed code in the only target that uses it -- NightscoutUploadKit. There may be some history here on why we want a unique target here for this, so just let me know. This also updates the code from Objc to Swift..

Note I am still using CommonCrypto. The other library CryptoKit can do the same thing here but I was trying to limit the amount of risk/change here.

When this PR merges, this NightscoutServiceKit PR will need merged as well https://github.com/ps2/NightscoutService/pull/29


ps2:
Thanks!

PR #760, opened 2023-01-07
Add remote actions to Nightscout client api

bill-foreflight:
This will add the following Remote 1.0 functionality to NigthscoutUploadKit to support Caregiver:

1. Remote override enacting
2. Remote bolus delivery
3. Remote carbs delivery

Caregiver is now using NigthscoutUploadKit for its other Nightscout client needs to (glucose, predicted loop values, device status, etc..)


MinimedKit/Extensions/NSDateFormatter.swift
```diff
@@ -20,4 +20,14 @@ extension DateFormatter {
 
         return formatter
     }
+
```
bill-foreflight: This is the same method copied from above, with the exception of the fractional seconds in the format. See notes at the bottom of the PR regarding this formatting addition.

NightscoutUploadKit/Models/COBStatus.swift
```diff
@@ -11,8 +11,8 @@ import Foundation
 public struct COBStatus {
     typealias RawValue = [String: Any]
```
bill-foreflight, line 13: Made some things public that I can use in Caregiver. Arguably I should just make all of these fields public, rather than partially like this.

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -8,22 +8,43 @@
...
 import Crypto
 
-public enum UploadError: Error {
+public enum UploadError: LocalizedError {
     case httpError(status: Int, body: String)
     case missingTimezone
     case invalidResponse(reason: String)
     case unauthorized
     case missingConfiguration
     case invalidParameters
```
bill-foreflight, line 15: This helps to describe the errors in Caregiver

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -277,8 +298,77 @@ public class NightscoutUploader {
             completion(.failure(UploadError.invalidParameters))
         }
     }
+    
+    
+    // MARK: - Loop Notifications
+
+    public func startOverride(overrideName: String, reasonDisplay: String, durationInMinutes: Int) async throws {
+
```
bill-foreflight, line 304: I did not make a model for these different payloads as I was not sure if it is worth it for this 1.0 remote functionality. Happy to if desired.

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -277,8 +298,69 @@ public class NightscoutUploader {
...
+        }
+        
+        try await postNotification(payload: jsonDict)
+    }
 
+    public func postNotification(payload: [String: String]) async throws  {
+        guard let url = url(for: .notifications) else {
+            throw UploadError.missingConfiguration
+        }
+
```
bill-foreflight, line 359: The notification API returns an "OK" string in the body. I _could_ check that but I feel the 200 response that gets checked in `postToNS` is sufficient.

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -483,6 +565,32 @@ public class NightscoutUploader {
         }
     }
```
bill-foreflight, line 565: This is not used yet but this can help us when ready to decode JSON from new concurrent APIs.

NightscoutUploadKit/NightscoutUploader.swift
```diff
@@ -483,6 +565,32 @@ public class NightscoutUploader {
...
     
+    func postToNS<T: Decodable>(_ json: Any, url: URL, jsonReturnType: T.Type?) async throws -> T {
+        let data = try await postToNS(json, url: url)
+        return try JSONDecoder().decode(T.self, from: data)
+    }
+
```
bill-foreflight, line 570: The base version returns Data - which can be useful if the endpoint doesn't return JSON (i.e. a String reponse)

NightscoutUploadKit/TimeFormat.swift
```diff
@@ -16,6 +16,11 @@ class TimeFormat: NSObject {
     }
```
bill-foreflight, line 18: I'm not sure why, but the `DeviceStatus.created_at`, coming from Nightscout, is not in the formatted `formatterISO8601`. This is odd as it appears that is what we are uploading. This unexpected format causes clients, like Caregiver, to fail to fetch the DeviceStatus when using NightscoutUploadKit. I think it may have something to do with some processing happening on Nightscout but I didn't dig in too much.

The change here is to fallback to the format that is coming from Nightscout.

NightscoutUploadKit/NightscoutUploader.swift
bill-foreflight: I am curious if any feedback on this error paradigm in general -- using LocalizedErrors and `throws`. The remote commands 2.0 will have the potential for deeply nested errors in Swift concurrent APIs, as there are many things that can cause a failure in distant code. My current approach has been to define Localized errors like this. Then most methods use `throws`, which gets propogated to the bottom of the callstack where I present whatever Error was thrown to the user -- without any knowledge of its type or where it is coming from. 

I mention this as I see a lot of Loop APIs that use Result Types and specialized error types for error handling -- that are not using Swift Concurrency. I think returning the Result type with specialized errors can work as well but the code seems cleaner/simpler when you can just place a `try` before failable calls, rather than thinking about what to do with each result (i.e. with guards, Result unwrapping, early returns, etc.)

When you see some of the remote 2.0 code, the question may make more sense and we can revisit it then if not clear.

MinimedKit/Extensions/NSDateFormatter.swift
ps2: I think you can do this: 
```
let formatter = ISO8601DateFormatter()
formatter.formatOptions =  [.withInternetDateTime, .withFractionalSeconds]
```

NightscoutUploadKit/NightscoutUploader.swift
ps2: I think this is fine. As long as the exposed methods are not expecting to be passed json dicts, or returning json dicts. They should implement whatever niceties they can to provide a "Swifty" API to users of the framework, following Swift idioms and conventions as much as makes sense.

Some of these are taking `durationInMinutes` or `absorptionInHours`.  While I like expressing the expected units of parameters, it might be better to just do a TimeInterval?

NightscoutUploadKit/NightscoutUploader.swift
ps2: The http status should override any body text in terms of failure semantics, I'd posit.  If NS is returning errors in responses with a 200 status, that'd be a bug in NS, I'd think.

ps2 (commented):
Looks good! I think maybe the formatter has a simple solution, though.

ps2:
Just FYI, none of these comments are blocking, or I would've marked the PR as changes requested. I just made the comments. If you want this to land as is, just let me know, or put it back up for review after you've made changes.

MinimedKit/Extensions/NSDateFormatter.swift
bill-foreflight: I updated this and the other method in this class to match since there was a TODO for the same. Since these are class methods on DateFormatter, I updated it to be an extension on ISO8601DateFormatter.

NightscoutUploadKit/NightscoutUploader.swift
bill-foreflight: Good idea on the TimeInterval. I updated both.

bill-foreflight:
@ps2 The updates should be done now - thanks for the suggestions.

NightscoutUploadKit/TimeFormat.swift
```diff
@@ -9,13 +9,18 @@
...
 
     static func dateFromTimestamp(_ string: String) -> Date? {
-        return formatterISO8601.date(from: string)
+        if let result = formatterISO8601.date(from: string) {
+            return result
+        }
+            
+        //Nightscout is returning this format in some cases...  needs more research
+        return ISO8601DateFormatter.fractionalSecondsFormatter().date(from: string)
```
ps2: Constructing formatters can be very expensive, involving locales, calendars, etc. Recommend to keep the construction of the formatter a singleton of sorts.

ps2 (changes requested):
Looks good, but I think this changed to constructing a formatter for each parsing, which can be expensive if it's used to decode things like arrays of data.

NightscoutUploadKit/TimeFormat.swift
gestrich: Good idea and normally I do use static but didn't think of it this time. I updated it to follow the paradigm in this class for the existing formatter here and made a private reference in this class.

PR #766, opened 2023-02-15
Display MDT battery

marionbarker:
This adds the pumpManager.status.pumpBatteryChargeRemaining to the Medtronic Pump Display as percent.

Also updated a few places where Text() instead of Text(LocalizedStrings()) was used.

MDT-display-with-battery-percent

### DIFF
--- a/MinimedKitUI/Views/MinimedPumpSettingsView.swift
+++ b/MinimedKitUI/Views/MinimedPumpSettingsView.swift
@@ -147,6 +147,15 @@ struct MinimedPumpSettingsView: View {
 
             Section() {
                 HStack {
+                    Text(LocalizedString("Pump Battery Remaining", comment: "Text for medtronic pump battery percent remaining")).foregroundColor(Color.primary)
+                    Spacer()
+                    if let chargeRemaining = viewModel.pumpManager.status.pumpBatteryChargeRemaining {
+                        Text(String("\(Int(round(chargeRemaining * 100)))%"))
+                    } else {
+                        Text(String(LocalizedString("unknown", comment: "Text to indicate battery percentage is unknown")))
+                    }
+                }
+                HStack {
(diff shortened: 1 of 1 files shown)


ps2:
Localization changes look good. Battery remaining isn't really "configuration".  Maybe it can go down by pump time?

marionbarker:
I love SwiftUI.
* I can put this anywhere.

If I used an MDT pump, I'd like to see the battery status higher up on the screen.
How about one of these options?

image171

BTW - I found a typo in one of my prior Localize lines and pushed the fix for it.

ps2:
Yeah, working in SwiftUI spoils you for UIKit chores, that's a big reason why I redid this UI. I think the battery belongs below the controls. This is one of those fields that we should strive to make it so the user doesn't need to worry about, and they can depend on Loop to tell them if they're getting close to replacement, or need to replace.

marionbarker:
Loop just told me to replace the battery now - said it was at 0% but pump was still working.

I've moved the location of the row and pushed the changed.

And with a fresh battery, the Pump icon on Nightscout has changed to nominal font (it was showing yellow last night).

MDT-battery-new-location

ps2:
Putting it in the same section as PumpTime would save some space. Does it work there?

marionbarker:
Option 1 - above pump time. Let me know if you want it below. (Graphic goes with commit 53f0f6e)

mdt-pump-battery-display

dsut4392:
I actually like where it's placed in your first screenshot. The argument that it's "one of those fields that we should strive to make it so the user doesn't need to worry about, and they can depend on Loop to tell them if they're getting close to replacement" applies at least as much to the reservoir volume, and that's right up the top in big bold font...

Even better would be to show it in the HUD, akin to how the HUD for pods shows both reservoir volume and remaining time, or just a tiny battery icon and % remaining like FAX does.

